PR #9, opened 2015-08-10
Feature/processing excerpt data

das-g:
Up to now without ESRI File Geodatabase support (see issue #10), as that'd require a [GDAL driver](http://www.gdal.org/drv_filegdb.html) relying on proprietary libraries.

Reviewed by:
- [ ] @njordan-hsr 
- [x] @wasabideveloper


docker-compose.yml
```diff
@@ -6,5 +6,13 @@ script:
     - PGHOST=db
   volumes:
     - tmp/osmosis/:/tmp/osmosis/
+excerpt:
+  build: osmaxx_excerpt/
+  links:
+    - db
```
wasabideveloper: Please use full names instead of shortcuts

docker-compose.yml
```diff
@@ -6,5 +6,13 @@ script:
     - PGHOST=db
   volumes:
     - tmp/osmosis/:/tmp/osmosis/
+excerpt:
```
wasabideveloper: I think this name is not clear enough about what the container is doing.

osmaxx_excerpt/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+    gdal-bin\
+    zip
+
+ENV HOME /home/excerpt
+
+WORKDIR $HOME
+
+ADD . $HOME
+
+CMD python excerpt.py 8.775449276 47.1892350573 8.8901920319 47.2413633153 -f spatialite
```
wasabideveloper: Please add a comment where these coordinates are. Switzerland?

docker-compose.yml
```diff
@@ -6,5 +6,13 @@ script:
...
+  build: osmaxx_excerpt/
+  links:
+    - db
+  environment:
+    - PGHOST=db
+  volumes:
+    - results:/home/excerpt/data/
 db:
```
wasabideveloper: Please use full names instead of shortcuts

osmaxx_excerpt/extract/extract_gpkg.sh
```diff
@@ -22,15 +22,21 @@ else
 	fi
 
 	echo "starting "$FILENAME
```
wasabideveloper: I'm a bit picky today. Starting on what?

osmaxx_excerpt/extract/extract_statistics
```diff
@@ -5,9 +5,9 @@ YMIN=$2
 XMAX=$3
 YMAX=$4
 CRS=900913
+mkdir -p `pwd`/tmp
 FILE=`pwd`/tmp/$5'_STATISTICS.csv'
```
wasabideveloper: Maybe we should move such hardcoded file names to settings?

osmaxx_excerpt/Dockerfile
wasabideveloper: How the container should be used (called) need to be documented.

PR #14, opened 2015-08-20
Dhruv's changes

das-g:
Reviewed by
- [ ] @sfkeller (SQL and SQL comments)
- [x] @njordan-hsr 
- [ ] @wasabideveloper 
- [x] @das-g

### DIFF
--- a/bootstrap/main-bootstrap.sh
+++ b/bootstrap/main-bootstrap.sh
@@ -66,19 +66,19 @@ echo "*** fill initial OSM data ***"
 PSQL='psql -v ON_ERROR_STOP=1 -U postgres '
 
 createfunctions(){
-  echo 'creating functions...'
+  echo 'CREATING FUNCTIONS'
   $PSQL -f ./src/create_functions.sql $DB_NAME
 }
 
 
 cleandata(){
-  echo 'cleaning database...'
+  echo 'CLEANING DATABASE'
   $PSQL -f ./src/sweeping_data.sql $DB_NAME
 }
 
 
 filterdata(){
-  echo 'filtering data...'
+  echo 'FILTERING DATA'
   sh filter_data.sh $DB_NAME
 }
 
@@ -94,4 +94,4 @@ create_statistics(){
 STARTTIME=$(date +%s)
 setup_db && init_osmosis  && fill_initial_osm_data && createfunctions && cleandata && filterdata && create_statistics
 ENDTIME=$(date +%s)
-echo "It took $(($ENDTIME - $STARTTIME)) seconds to complete..."
+echo "It took $(($ENDTIME - $STARTTIME)) seconds to complete."
--- a/bootstrap/src/transfer_sql/address.sql
+++ b/bootstrap/src/transfer_sql/address.sql
@@ -30,7 +30,7 @@ CREATE TABLE osmaxx.address_p(
 INSERT INTO osmaxx.address_p
   SELECT osm_id as osm_id,
 	osm_timestamp as lastchange,
-	'N' AS geomtype, -- When address is linked to a node -- 
+	'N' AS geomtype, -- When address is linked to a node 
 	way AS geom,
 	'b' AS type, --When address is linked to a building
 	name as name,
@@ -56,8 +56,8 @@ UNION
   SELECT osm_id as osm_id,
 	osm_timestamp as lastchange,
 	CASE 
-	 WHEN osm_id<0 THEN 'R' -- R=Relation --
-	 ELSE 'W' 		-- W=Way --
+	 WHEN osm_id<0 THEN 'R' -- R=Relation 
+	 ELSE 'W' 		-- W=Way 
 	 END AS geomtype, 
 	ST_Centroid(way) AS geom,
 	'b' AS type,
@@ -87,7 +87,7 @@ where building not in ('entrance') and ("addr:street" is not null or "addr:place
 INSERT INTO osmaxx.address_p
   SELECT osm_id as osm_id,
 	osm_timestamp as lastchange,
-	'N' AS geomtype, 	-- Node --
+	'N' AS geomtype, 	-- Node 
 	way AS geom,
 	'e' AS type,
 	name as name,
@@ -113,8 +113,8 @@ INSERT INTO osmaxx.address_p
   SELECT osm_id as osm_id,
 	osm_timestamp as lastchange,
 	CASE 
-	 WHEN osm_id<0 THEN 'R' -- R=Relation --
-	 ELSE 'W' 		-- W=Way --
+	 WHEN osm_id<0 THEN 'R' -- R=Relation 
+	 ELSE 'W' 		-- W=Way
 	 END AS geomtype, 
 	ST_Centroid(way) AS geom,
 	'e' AS type,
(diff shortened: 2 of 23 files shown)


bootstrap/main-bootstrap.sh
```diff
@@ -57,7 +45,7 @@ init_osmosis() {
 
 fill_initial_osm_data(){
 echo "*** fill initial OSM data ***"
-    wget -nc -q --progress=bar http://download.geofabrik.de/europe/switzerland-latest.osm.pbf -O  $WORKDIR_OSM/switzerland-latest.osm.pbf
+    wget -q --progress=bar http://download.geofabrik.de/europe/azores-latest.osm.pbf -O $WORKDIR_OSM/switzerland-latest.osm.pbf
```
njordan-hsr: azores doesn't sound like switzerland to me ;)

extract/excerpt.py
```diff
@@ -18,57 +20,17 @@ def to_mercator(mercator_x_lon, mercator_y_lat):
...
-
-def call_fgdb(data, name):
-    get_statistics(data, name)
-    gdbcmd = 'sh', './extract/extract_fgdb.sh', data[0], data[1], data[2], data[3], name
-    subprocess.check_call(gdbcmd)
-
-
-def call_shp(data, name):
+#Call the shell script that creates the database in different formats
+def call_db(data, name, form):
```
njordan-hsr: I don't see any database created here, even though the comment says so. Is this not needed here anymore?

README.md
```diff
@@ -28,7 +28,7 @@ To extract an excerpt, you need to do the following:
 * Prepare the database (will download Switzerland)
 ```shell
 docker-compose run bootstrap /bin/bash -c \
-  'sleep 10 && sh main-bootstrap.sh {excerptSouthBorder} {excerptWestBorder} {excerptNorthBorder} {excerptEastBorder}'
+  'sleep 10 && sh main-bootstrap.sh {excerptWestBorder} {excerptSouthBorder} {excerptEastBorder}  {excerptNorthBorder}'
```
njordan-hsr: Where did the ordering of the borders change? Or was the code already doing it this way and was just documented incorrectly?

bootstrap/main-bootstrap.sh
```diff
@@ -37,7 +26,6 @@ setup_db() {
     createdb   -U postgres $DB_NAME
     execute_sql "CREATE EXTENSION hstore;"
     execute_sql "CREATE EXTENSION postgis;"
-    execute_sql "CREATE EXTENSION postgis_topology;"
```
njordan-hsr: Are we sure we don't need that in the cleaning up process and the like? Don't know the code well enough to be sure...

ie. something like outlined here: [clean up road networks](http://blog.mathieu-leplatre.info/use-postgis-topologies-to-clean-up-road-networks.html)

bootstrap/filter_data.sh
```diff
@@ -14,100 +14,110 @@ if [ "$1" != "" ]; then
 
   $PSQL -c "CREATE SCHEMA osmaxx;" $DATABASE
 
-  echo 'filtering..'
+  echo 'filtering address..'
   STARTTIME=$(date +%s)
   mkdir -p tmp  # TODO: Should /dev/null be used instead? Or a non-interactive pager?
   $PSQL -f $DIR/src/transfer_sql/address.sql $DATABASE
   ENDTIME=$(date +%s)
   echo "It takes $(($ENDTIME - $STARTTIME)) seconds to complete this task..."
```
das-g: Please correct this (and other occurances) to past tense ('took' instead of 'takes').

bootstrap/filter_data.sh
```diff
@@ -14,100 +14,110 @@ if [ "$1" != "" ]; then
 
   $PSQL -c "CREATE SCHEMA osmaxx;" $DATABASE
 
-  echo 'filtering..'
+  echo 'filtering address..'
```
das-g: :+1:Good idea to give a bit more feedback here. :smiley:

extract/excerpt.py
das-g: Indeed, this does not create the/a database. It exports files of the specified format from an existing database. @Dhruv222  Please correct the comment accordingly.

Also, please use a parameter name with 'format' in it (e.g. `file_format`), as 'form' is a different English word with a quite different meaning.

extract/excerpt_help.md
```diff
@@ -0,0 +1,9 @@
+# excerpt.py documentation
+
+## functions
+
+1. name_generator- generates the name of the db depending on the date and time
+2. to_mercator- converts the map projection of the default osm format to the mercator projection
+3. get_statistics- calls the extract_statistics.sh with the coordinates which calculates the statistics for csv file
```
das-g: This file is currently only called `extract_statistics`, not `extract_statistics.sh`. Let's keep the documentation like this and rename the file with #15, though.

extract/excerpt.py
das-g: The name `call_db` is not very indicative of what this function does.

bootstrap/src/transfer_sql/address.sql
```diff
@@ -25,12 +25,14 @@ CREATE TABLE osmaxx.address_p(
 ------------------
 --   building   --
 ------------------
+
+-- Without the Entrance Node i.e; just the building --
 INSERT INTO osmaxx.address_p
   SELECT osm_id as osm_id,
 	osm_timestamp as lastchange,
-	'N' AS geomtype, 
+	'N' AS geomtype, -- When address is linked to a node --
```
das-g: While an SQL comment is started with a double dash (`--`), it doesn't need to be terminated by another double dash. It just continues until the end of the line.

While it makes sense to enclose a comment in double dashes to make a title (like

      ------------------ 
      --   building   -- 
      ------------------ 
above), please do not end inline comments with a double dash.

bootstrap/main-bootstrap.sh
Dhruv222: i was testing different regions which is why azores is there, I will change it.

Also is -nc is not removed and the link is changed, for example from switzerland to azores, the database being used will still be switzerland resulting an empty database as the excerpt would lie in azores.

bootstrap/filter_data.sh
Dhruv222: Will do.

bootstrap/main-bootstrap.sh
Dhruv222: Prof Stefan asked me to remove the postgis topology extension as it was not being used, but if it is needed i will add it. As I am not sure where it is used.

extract/excerpt.py
Dhruv222: i have changed the parameter name and for the function is export_format okay?

bootstrap/main-bootstrap.sh
das-g: If unsure, leave it out as Stefan said. If it's needed, we'll notice soon enough.

extract/excerpt.py
```diff
@@ -18,57 +20,19 @@ def to_mercator(mercator_x_lon, mercator_y_lat):
...
-
-def call_fgdb(data, name):
-    get_statistics(data, name)
-    gdbcmd = 'sh', './extract/extract_fgdb.sh', data[0], data[1], data[2], data[3], name
-    subprocess.check_call(gdbcmd)
-
-
-def call_shp(data, name):
+#Calls the shell script that exports files of the specified format(file_format) from existing database
+def export_format(data, name, file_format):
```
das-g: The name `export_format` is confusing, too, as this might be interpreted that this function would return the export format (i.e. the format used for the export), when it actually performs the export. Suggestions: `export_to_format` or `export_from_db` or even `export_from_db_to_format`

Dhruv222: Okay I will use the third one.

PR #33, opened 2015-10-26
Feature/merge bootstrap with extraction

njordan-hsr:
Merge bootstrapping and extracting into one container and make it work again similarly as before.

Implements parts of [#243](https://github.com/geometalab/osmaxx/issues/243)

PR #34, opened 2015-10-30
Feature/243 implement post gis worker

njordan-hsr:
Is based upon #33 

Implements a (scalable) worker for all required formats using RQ.

Includes the working database inside the container.

Closes geometalab/osmaxx#243


README.md
```diff
@@ -18,47 +18,38 @@ docker-compose build
...
-docker-compose stop --timeout 0 # Only use --timeout 0 if you remove the container after
-docker-compose rm -f
+docker-compose run --rm worker /bin/bash -c \
+   'python3 worker/converter_job.py \
+   --west 8.775449276 --south 47.1892350573 --east 8.8901920319 --north 47.2413633153\
+   -f fgdb -f shp -f gpkg -f spatialite'
 ```
 
-Further Documentation can be found under `docs/`.
+The same in one line for easier copy paste:
```
das-g: Do some terminals have issues with pasted multi-line commands?

worker/00_make_db_utf8.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# make it utf8!
+if ! $(psql template1 -c 'SHOW SERVER_ENCODING' | grep -q UTF8); then
+  psql postgres -c "update pg_database set datallowconn = TRUE where datname = 'template0';"
+  psql template0 -c "update pg_database set datistemplate = FALSE where datname = 'template1';"
+  psql template0 -c "drop database template1;"
+  psql template0 -c "create database template1 with template = template0 encoding = 'UTF8';"
+  psql template0 -c "update pg_database set datistemplate = TRUE where datname = 'template1';"
+  psql template1 -c "update pg_database set datallowconn = FALSE where datname = 'template0';"
+fi
```
das-g: :white_check_mark: created issue to follow up on questions we have about this: geometalab/osmaxx#251

README.md
njordan-hsr: No, but I did use it quite often, so I thought what helps me, might help others as well. The issue with multi-line commands is that (at least in my terminal) it has problems with the history.

worker/converter_job.py
```diff
@@ -1,35 +1,77 @@
...
+            set_status_on_job(JobStatus.ERROR)
+            self.notify()
+            raise
+
+    def notify(self):
+        self._notify_status_change()
+
+    def _notify_status_change(self):
+        """
+        fire and forget, and don't care when exceptions occur.
```
das-g: Should we care about some exceptions caused by the `GET` request? Should we at least log them?

njordan-hsr: No, since we don't really require to send a callback url just yet, see also: https://github.com/geometalab/osmaxx-postgis-conversion/blob/feature/244_conversion_job_manager/conversion_service/worker/converter_job.py#L53.
Should probably be done better there.

PR #40, opened 2015-11-10
Feature/244 conversion job manager

hixi:
conversion job manager and some tests for it.

```bash
./runtests.py # success
tox # success
```

Closes geometalab/osmaxx#244


conversion_service/conversion_job/models.py
```diff
@@ -0,0 +1,130 @@
...
+    def get_resulting_file_path_or_none(self, format):
+        """
+        :param format: format of the file
+        :return: path to result file
+        """
+        file_paths = []
+        for name in os.listdir(self.output_directory):
+            result_file_path = os.path.join(self.output_directory, name)
+            if os.path.isfile(result_file_path):
+                file_paths.append(result_file_path)
```
das-g: The above 6 lines get the paths of all regular files (directly) in the directory `self.output_directory`. We should make this a utility function with the directory as argument.

conversion_service/conversion_job/models.py
```diff
@@ -0,0 +1,130 @@
...
+class GISFormat(models.Model):
+    conversion_job = models.ForeignKey(ConversionJob, verbose_name=_('conversion job'), related_name='gis_formats')
+    format = models.CharField(_('format'), choices=CONVERTER_CHOICES['output_formats'], max_length=10)
+    progress = models.IntegerField(
+        _('progress'),
+        choices=ConversionProgress.choices(),
+        default=ConversionProgress.NEW.value,
+    )
+
+    def get_result_file_path(self):  # pragma: no cover
```
das-g: What's the reason for not including this in coverage?

hixi: Because it is only calling a already tested function (from above). I don't think testing if it calls that function is a very useful test.

conversion_service/conversion_job/models.py
das-g: :white_check_mark: done with 946804b

conversion_service/conversion_job/models.py
```diff
@@ -0,0 +1,126 @@
...
+        :return: path to result file
+        """
+        file_paths = get_file_only_path_list_in_directory(self.output_directory)
+        if not file_paths:
+            return None
+
+        file_path = [path for path in file_paths if path.split('.zip')[:-1][0].endswith(format)]
+        if not file_path:
+            return None
+        return file_path[0]
```
das-g: This would be much easier if we had separate output dirs for each ordered format of a job.

hixi: What would that make easier? We still would need to check if the file is there or not, and also check if the directory exists before that.

hixi:
```python
file_dir = os.path.join(self.output_directory, format)
        if not os.path.exists(file_path):
            return None
file_paths = get_file_only_path_list_in_directory(file_dir)

file_path = [path for path in file_paths if path.split('.zip')[:-1][0].endswith(format)]
if not file_path:
    return None
return file_path[0]
```

conversion_service/conversion_job/serializers.py
```diff
@@ -0,0 +1,130 @@
...
+    def create(self, validated_data):  # pragma: nocover -> is being tested through ConversionJobSerializer
+        formats = [GISFormat(**item) for item in validated_data]
+        return GISFormat.objects.bulk_create(formats)
+
+    def to_representation(self, data):
+        return data.values_list('format', flat=True)  # pragma: nocover -> being tested through ConversionJobSerializer
+
+    def to_internal_value(self, data):
+        """
+        List of strings to list of dicts of native values <- List of dicts of primitive datatypes.
```
das-g: How should this Docstring be understood? I guess left of the `<-` is the return value, right of the `<-` the input. But what is a "List of strings to list of dicts of native values"?

conversion_service/conversion_job/serializers.py
```diff
@@ -0,0 +1,124 @@
...
+
+class GISFormatListSerializer(serializers.ListSerializer):
+    def to_representation(self, data):
+        return data.values_list('format', flat=True)
+
+    def to_internal_value(self, data):
+        """
+        List of strings to list of dicts of native values <- List of dicts of primitive datatypes.
+        """
+        ret = [{'format': value} for value in data]
```
das-g: Other than the variable name suggests, `ret` is not the return value. Please choose a better name or inline it.

conversion_service/conversion_job/serializers.py
```diff
@@ -0,0 +1,124 @@
...
+        fields = ('crs', 'detail_level',)
+
+
+class ConversionJobSerializer(serializers.ModelSerializer):
+    extent = ExtentSerializer(many=False)
+    gis_formats = GISFormatSerializer(required=False, many=True)
+    gis_option = GISOptionSerializer()
+    status = StatusHyperlinkSerializer(source='rq_job_id', read_only=True)
+
+    def create(self, validated_data):
```
das-g: Should this function have side effects on the passed in `validated_data`, or should it only modify an internal copy of that dict?

conversion_service/conversion_job/serializers.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class GISOptionSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = GISOption
+        fields = ('crs', 'detail_level',)
+
+
+class ConversionJobSerializer(serializers.ModelSerializer):
+    extent = ExtentSerializer(many=False)
```
das-g: Isn't `False` the default for `many`?

conversion_service/conversion_job/serializers.py
hixi: since we're not returning the validate_data and not doing anything with it, e.g. using super(), I don't see why this should be necessary.

Achieving it would be trivial, though:

```python
validated_data = validated_data.copy()
...
```

conversion_service/conversion_job/serializers.py
hixi: Yes, it is, but since it is a ForeignKey (with limit = 1 on both sides), I wanted to make it more explicit.

conversion_service/conversion_job/serializers.py
das-g: The function shouldn't make assumptions about whether the caller will or will not use the passed object after the call returns. So unless its purpose is to modify the passed object, it shouldn't have side effects on it.

conversion_service/conversion_job/serializers.py
das-g: :white_check_mark: keep as-is, then

conversion_service/conversion_job/serializers.py
das-g: :white_check_mark: inlined with af476a7

conversion_service/conversion_job/serializers.py
```diff
@@ -0,0 +1,121 @@
...
+    class Meta:
+        model = ConversionJob
+        fields = ('id', 'rq_job_id', 'callback_url', 'status', 'gis_formats', 'gis_options', 'extent')
+        depth = 1
+        read_only_fields = ('rq_job_id', 'status',)
+
+
+# Status-only serializers
+
+class DownloadURL(serializers.HyperlinkedRelatedField):
```
das-g: Most serializer field class names seem to end in `Field`. Should this one be called `DownloadURLField`?

conversion_service/conversion_job/views.py
```diff
@@ -0,0 +1,76 @@
...
+
+class ExtentViewSet(viewsets.ModelViewSet):
+    """
+    This viewset automatically provides `list`, `create`, `retrieve`,
+    `update` and `destroy` actions.
+    """
+    queryset = Extent.objects.all()
+    serializer_class = ExtentSerializer
+    permission_classes = (
+        permissions.IsAuthenticatedOrReadOnly,
```
das-g: Does this allow unauthenticated reading?

conversion_service/conversion_job/views.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def _update_status_from_rq(self):
+        conversion_job = self.get_object()
+        rq_job = django_rq.get_queue().fetch_job(job_id=conversion_job.rq_job_id)
+
+        # only do work if the job is not yet deleted
+        if rq_job:
+            # don't set the status if we're already done
+            if conversion_job.status != JobStatus.DONE.value:
+                conversion_job.status = rq_job_status_mapping[rq_job.status].value
```
das-g: Is the condition needed? Can `conversion_job.status` ever be `DONE.value` when `rq_job.status` isn't `RQJobStatus.FINISHED`?

If we think it cannot and must not, shouldn't we fail with an exception if this combination occurs anyway? (And shouldn't the model be responsible of ensuring that, rather than a view?)

conversion_service/conversion_job/views.py
```diff
@@ -0,0 +1,76 @@
...
+
+            progress = rq_job.meta.get('progress', None)
+            if progress:
+                progress_state = progress.value
+                for gis_format in conversion_job.gis_formats.all():
+                    if gis_format.progress != ConversionProgress.SUCCESSFUL.value:
+                        gis_format.progress = progress_state
+                        gis_format.save()
```
das-g: Is the `gis_format.progress != ConversionProgress.SUCCESSFUL.value` condition needed? Can `gis_format.progress` ever be `SUCCESSFUL.value` when `progress.value` isn't?

If we think it cannot and must not, shouldn't we fail with an exception if this combination occurs anyway? (And shouldn't the model be responsible of ensuring that, rather than a view?)

conversion_service/conversion_job/views.py
```diff
@@ -0,0 +1,76 @@
...
+            conversion_job.save()
+
+
+class GISFormatStatusViewSet(viewsets.mixins.RetrieveModelMixin, viewsets.GenericViewSet):
+    queryset = GISFormat.objects.all()
+    serializer_class = GISFormatStatusSerializer
+
+    @detail_route()
+    def download_result(self, *args, **kwargs):  # pragma: nocover
```
das-g: This function shouldn't be excluded from coverage, I think.

conversion_service/converters/converter_settings.py
```diff
@@ -2,8 +2,7 @@
 
 OSMAXX_CONVERSION_SERVICE = {
     'XAPI_MIRROR': 'http://www.overpass-api.de/api/xapi_meta',
-    'WORKDIR': '/tmp/osm_converter',
-    'RESULTDIR': '/tmp/results',
+    'RESULT_DIR': '/tmp/osm/results',
```
das-g: Why `osm`? Do we have non-OSM results, too?

conversion_service/conversion_job/views.py
hixi: [Yes](http://www.django-rest-framework.org/api-guide/permissions/#isauthenticatedorreadonly)

conversion_service/conversion_job/views.py
hixi: This does not what you think: if the status is set to `JobStatus.DONE`, it shouldn't be changed anymore.

conversion_service/conversion_job/views.py
hixi: Same as above.

conversion_service/converters/converter_settings.py
hixi: we might not want to assume there is no /tmp/results dir already on tmp, in order to get a better overview when looking into the /tmp/ folder.
In case of using the docker container, this location is changed to [/data/results](https://github.com/geometalab/osmaxx-postgis-conversion/pull/40/files#diff-5f8ab60a6ba7194f5bbcae419d7a2b3eR22)

conversion_service/conversion_job/views.py
hixi: care to elaborate on why? I don't think this needs test as it stands currently:
`self.get_object().get_result_file_path()` is a function calling another function that is already being tested, and then we would more or less only test the third party/library functionality of the response object.

das-g: This function does non-trivial stuff beyond calling `self.get_object().get_result_file_path()`. (_Nearly_ trivial, granted, but still non-trivial enough to potentially be buggy.) I'm not saying we must test it, but if we aren't testing it, coverage should tell us that we aren't.

To test it (if we want to do that) without re-testing `get_result_file_path`, mock the latter. One would probably also want to mock `open`.

conversion_service/conversion_job/serializers.py
hixi: Yes, it should.

conversion_service/conversion_job/models.py
das-g: :white_check_mark: since 2c3f2da, not excluded anymore

conversion_service/conversion_job/models.py
das-g: `s/much easier/less messy/`

It wouldn't have to make assumptions about the file name structure.

conversion_service/conversion_job/serializers.py
```diff
@@ -0,0 +1,121 @@
...
+
+class GISFormatListSerializer(serializers.ListSerializer):
+    def to_representation(self, data):
+        return data.values_list('format', flat=True)
+
+    def to_internal_value(self, data):
+        """
+        List of strings to list of dicts of native values <- List of dicts of primitive datatypes.
+        """
+        return super().to_internal_value([{'format': value} for value in data])
```
das-g: How should this Docstring be understood? I guess left of the `<-` is the return value, right of the `<-` the input. But what is a "List of strings to list of dicts of native values"?

conversion_service/conversion_job/models.py
hixi: Since this introduces quite a few changes, I propose to postpone this. I created an issue for it:
geometalab/osmaxx#258

conversion_service/manager/job_manager.py
```diff
@@ -0,0 +1,23 @@
+from manager.rq_helper import rq_enqueue_with_settings
+from worker.converter_job import convert
+
+
+class ConversionJobManager:
+    """
+
+    :param geometry:
+    :param format_options:
+    """
```
das-g: useless Dockstring (only repeats `__init__` signature)

conversion_service/conversion_job/models.py
das-g: :white_check_mark:

runtests.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#! /usr/bin/env python
```
das-g: Looks like a previous change got lost here. (Maybe during merge conflict resolution.)

runtests.py
```diff
@@ -12,7 +12,7 @@
     'fast': ['tests', '-q'],
 }
 
-MODULES_TO_LINT = ['rest_api', 'tests', 'manager', 'worker', 'converters', 'utils']
+MODULES_TO_LINT = ['conversion_service', 'tests', ]
```
das-g: Should we additionally also lint `runtests.py` and `setup.py`?

hixi: have you actually tried it? It is clean (other than line-length, which is excluded anyway) - so why not :-)

tests/conversion_job/model_test.py
```diff
@@ -0,0 +1,194 @@
...
+        e = Extent(west=0, south=0, east=0, north=0)
+        # no exception expected
+        e.save()
+
+
+class ConversionJobTest(TestCase):
+    def test_output_directory_is_created_if_not_exists(self):
+        extent = Extent.objects.create(west=0, south=0, east=0, north=0)
+        conversion_job = ConversionJob(extent=extent)
+        conversion_job.save()
```
das-g: These 3 lines are repeated for every test method in this TestCase. Should they be moved into the `setUp` (with the `conversion_job` as instance variable)?

runtests.py
das-g: :white_check_mark: files included with 23055a6

No, I hadn't tried it.

tests/conversion_job/model_test.py
hixi: good idea.

tests/conversion_job/model_test.py
```diff
@@ -0,0 +1,182 @@
...
+    def test_get_saved_conversions_options(self):
+        extent = Extent.objects.create(west=0, south=0, east=0, north=0)
+        conversion_job = ConversionJob(extent=extent)
+        conversion_job.save()
+        format_choice = converter_options.get_output_formats()[2]
+        gis_format = GISFormat.objects.create(conversion_job=conversion_job, format=format_choice)
+        self.assertEqual(
+            conversion_job.get_conversion_options().get_output_formats()[0],
+            gis_format.format
+        )
```
das-g: As simple as it seems to be, this test confuses me. Can't quite say why.

tests/conversion_job/model_test.py
```diff
@@ -0,0 +1,166 @@
...
+        os.rmdir(directory)
+        self.assertFalse(os.path.exists(directory))
+        directory = self.conversion_job.output_directory
+        self.assertTrue(os.path.exists(directory))
+        self.assertEqual(
+            directory,
+            os.path.join(converter_settings.OSMAXX_CONVERSION_SERVICE['RESULT_DIR'], str(self.conversion_job.id)))
+        os.rmdir(directory)
+
+    def test_get_resulting_file_path_or_none_returns_none_with_unavailable_formats(self):
```
das-g: There is a popular test method naming convention that is almost, but not completely, followed in our test cases:

    test_<action>[_with_<parameters>][_when_<condition>]_<expected effect>(...)
For the example in the line above, this would man a name like

    def test_get_resulting_file_path_or_none_with_unavailable_format_returns_none(self):

Should we rename our test methods to completely follow this convention? (Or are they already following a different convention? It's very possible that the one I know isn't the only popular one around.)

tests/conversion_job/model_test.py
```diff
@@ -0,0 +1,166 @@
...
+            self.assertEqual(
+                self.conversion_job.get_resulting_file_path_or_none(format=conversion_format),
+                file_path
+            )
+        finally:
+            # cleanup
+            os.unlink(file_path)
+            os.rmdir(out_dir)
+
+    def test_get_resulting_file_path_or_none_returns_none_with_unavailable_format(self):
```
das-g: What's the difference to `test_get_resulting_file_path_or_none_returns_none_with_unavailable_formats` above?

tests/conversion_job/serializers_test.py
```diff
@@ -0,0 +1,109 @@
...
+        self.format_status_serializer = GISFormatStatusSerializer(self.gis_format, context={'request': request})
+
+    def tearDown(self):
+        shutil.rmtree(self.conversion_job.output_directory)
+        super().tearDown()
+
+    def _create_valid_file(self):
+        matching_file_names = ['{}.zip'.format(f) for f in converter_options.get_output_formats()]
+        for matching_file_name in matching_file_names:
+            open(os.path.join(self.conversion_job.output_directory, matching_file_name), 'x').close()
```
das-g: This method creates multiple files, doesn't it? If so, the name should reflect that.

tests/conversion_job/serializers_test.py
```diff
@@ -0,0 +1,109 @@
...
+        self.gis_format.save()
+        self.assertIsNone(self.format_status_serializer.data.get('result_url'))
+
+    def test_get_download_url_is_avilable_if_file_is_avaiable_even_if_progress_is_not_success(self):
+        self._create_valid_file()
+        os.path.join(self.conversion_job.output_directory)
+        self.assertEqual(
+            'http://some-host/gis_format_status/1/download_result/',
+            self.format_status_serializer.data.get('result_url')
+        )
```
das-g: Will we rely on this behaviour? If not, we could leave it unspecified (i.e. untested).

tests/conversion_job/view_test.py
```diff
@@ -0,0 +1,62 @@
...
+    def fetch_job(*args, **kwargs):
+        rq_job_mock = Empty()
+        rq_job_mock.status = RQJobStatus.STARTED
+        rq_job_mock.meta = {'progress': ConversionProgress.STARTED}
+        return rq_job_mock
+
+
+def django_rq_get_queue_mock():
+    django_rq_get_queue = DjangoRQGetQueueMock()
+    return django_rq_get_queue
```
das-g: If we stick to the [nomenclature of Martin Fowler](http://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs) and [xUnit Patterns.com](http://xunitpatterns.com/Mocks,%20Fakes,%20Stubs%20and%20Dummies.html), all of the above (`DjangoRQGetQueueMock`, `rq_job_mock`, `django_rq_get_queue_mock`) are **'stubs'** (test doubles providing pre-canned answers), **not 'mocks'** (test doubles checking or recording behaviour of their users, usually the system under test).

tests/conversion_job/view_test.py
```diff
@@ -0,0 +1,62 @@
...
+            ConversionProgress.NEW.value
+        )
+
+        conversion_job_status_view = ConversionJobStatusViewSet(kwargs={'rq_job_id': self.conversion_job.rq_job_id})
+        conversion_job_status_view._update_status_from_rq()
+
+        self.assertEqual(
+            min(conversion_job.gis_formats.values_list('progress', flat=True)),
+            ConversionProgress.STARTED.value
+        )
```
das-g: I think this test tells several stories and should be split up into several test methods.

tests/utils/directory_helper_test.py
```diff
@@ -0,0 +1,66 @@
...
+    def _assert_list_equal_file_names(self, dirs=(), file_names=()):
+        self._create_dirs_and_files_in_dir(dirs=dirs, file_names=file_names)
+        path_list_from_function = get_file_only_path_list_in_directory(self.random_directory)
+        result_list = [os.path.join(self.random_directory, file_name) for file_name in file_names]
+        path_list_from_function.sort()
+        result_list.sort()
+        self.assertListEqual(
+            path_list_from_function,
+            result_list
+        )
```
das-g: Test methods usually perform three phases (any of which can by a NoOp), known as the **3 'A's'**:
* **A**rrange (preparation - set up of environment, test doubles and starting state of the system under test)
* **A**ct (run the system under test)
* **A**ssert (check the expectations about the effect and return value of running the system under test)

While the above helper method might extract stuff common to all test methods that use it, it spans multiple of those phases (all 3 even, I think) and thus makes the test methods harder to understand when reading them with the "3 'A's" structure in mind.

tests/utils/directory_helper_test.py
```diff
@@ -0,0 +1,66 @@
...
+        self._assert_list_equal_file_names(dirs=directories)
+
+    def test_get_file_only_path_list_in_directory_does_not_recurse(self):
+        directories = ['testdir1', 'directory2', '_some_other_dir']
+        self._create_dirs_and_files_in_dir(dirs=directories, file_names=[])
+        file_names_in_subdir = ['test1.test', 'test2.txt', '_no_extension_test3']
+        for directory in directories:
+            for file_name in file_names_in_subdir:
+                open(os.path.join(self.random_directory, directory, file_name), 'x').close()
+        self.assertListEqual(get_file_only_path_list_in_directory(self.random_directory), [])
```
das-g: :+1: Nice explanatory test

conversion_service/conversion_job/serializers.py
das-g: :white_check_mark: changed with a65ba3a

conversion_service/conversion_job/views.py
das-g: :white_check_mark: changed with 49c2ad49550630232ea715f61930822150e7351b

conversion_service/conversion_job/views.py
das-g: Discussed with hixi: As far as we know, the status & progress on the RQ job won't change anymore when it has reached `DONE` or `SUCCESSFUL`. Assigning the same status again to the job DB model won't have any effect for the saving. (Save only acts on changed values.)

Ensuring that no assignments lead to unwanted state changes (i.e., that the above assumptions are correct) isn't the serializer's job. It would be the model's job, but we won't implement any checking for this in the model for now. (Might or might not do that later.)

:white_check_mark: simplified with 7f13c68f41c58ced6c2a6fe25283c3287eadedc4

conversion_service/conversion_job/views.py
das-g: :white_check_mark: see above (https://github.com/geometalab/osmaxx-postgis-conversion/pull/40#discussion_r44793251)

conversion_service/conversion_job/views.py
das-g: :white_check_mark: included again in coverage with ffd6494b1213d6ab25121b160a8e0bf4c92eb444

conversion_service/converters/converter_settings.py
das-g: :white_check_mark: changed to `/tmp/osmaxx-conversion-service/results` with 05b869c

conversion_service/conversion_job/serializers.py
das-g: :white_check_mark: removed with 7113c5104169c647329ba2f113e989baea74d4b6

conversion_service/manager/job_manager.py
das-g: :white_check_mark: removed with fa166ab4789c44a390e5907b96977dd0f1356bcc

runtests.py
das-g: :white_check_mark: fixed with 7e4235eb8316f57df833e0f8e973fc3ee3a0caaa

tests/conversion_job/model_test.py
das-g: :white_check_mark: done with  b770b0d

conversion_service/conversion_job/serializers.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class ConversionJobSerializer(serializers.ModelSerializer):
+    extent = ExtentSerializer(many=False)
+    gis_formats = GISFormatSerializer(required=False, many=True)
+    gis_options = GISOptionSerializer()
+    status = StatusHyperlinkSerializer(source='rq_job_id', read_only=True)
+
+    def create(self, validated_data):
+        gis_formats = validated_data.pop('gis_formats')
```
das-g: Should or should we not perform the modifications on a copy `validated_data` instead of on the original? (See comments at https://github.com/geometalab/osmaxx-postgis-conversion/pull/40#discussion-diff-44516445)

Is there an explicit contract with the caller (django rest framework and maybe our own future code) that the passed dict may be modified? (E.g. stated in the django rest framework documentation) If not, I'd vote for copying, even if the current implementation of django rest framework doesn't rely on the passed dict to remain unmodified.

If we copy, should we make a [shallow](https://docs.python.org/3/library/copy.html#copy.copy) or [deep copy](https://docs.python.org/3/library/copy.html#copy.deepcopy)? As we currently only assign to the top level of the dict, a shallow one would suffice. But a deep copy would preserve the clean behaviour even if that changes.

das-g:
I'm through with reviewing the changes of this pull request. Except for the issues & topics stated in the line comments, I don't have any objections.

tests/conversion_job/model_test.py
hixi: there is an (existing) sample file, which therefore can be returned, above, there is none.

tests/conversion_job/model_test.py
hixi: I'm quite certain, that there are others - but TBH, I did not try very hard to follow any. Renaming them makes sense to me.

tests/conversion_job/serializers_test.py
hixi: We don't, I'll remove it.

tests/conversion_job/view_test.py
hixi: :white_check_mark:

conversion_service/conversion_job/serializers.py
hixi: In the [DRF-Documentation](http://www.django-rest-framework.org/api-guide/serializers/#writing-create-methods-for-nested-representations) they do exactly that.

* `create` is the last method on a chain, which 'consumes' that validated_data, ie it is not used anymore later on

If you had looked at the docstring for `create`, it would have told you already that `pop` is considered OK.

```
...
        If there are many to many fields present on the instance then they
        cannot be set until the model is instantiated, in which case the
        implementation is like so:

            example_relationship = validated_data.pop('example_relationship')
            instance = ExampleModel.objects.create(**validated_data)
            instance.example_relationship = example_relationship
            return instance
...
```

tests/conversion_job/model_test.py
das-g: :white_check_mark: tests got more distinctive names with 6d310b5e403684644b0456228f69849c5f27095c

conversion_service/conversion_job/serializers.py
das-g: I can't see where in the docu they say that it's the last in the chain or that `validated_data` won't be used afterwards. But I can indeed see in the code that this is the case I can see the `pop`s suggested in the documentation and [in the Docstring](https://github.com/tomchristie/django-rest-framework/blob/54b79db2142675b8005e573b7af28c097ef7f72d/rest_framework/serializers.py#L823), so this should be fine.

:white_check_mark: leave as-is (without copy)

conversion_service/conversion_job/serializers.py
das-g: :white_check_mark: leave as-is, see https://github.com/geometalab/osmaxx-postgis-conversion/pull/40#discussion_r44858509

tests/conversion_job/serializers_test.py
das-g: removed in c51bfc878890161f83cdabf282120a1d673a4a00

tests/conversion_job/view_test.py
```diff
@@ -0,0 +1,67 @@
...
+            format=converter_options.get_output_formats()[3]
+        )
+
+    @patch('django_rq.get_queue', django_rq_get_queue_stub)
+    @patch('conversion_job.views.ConversionJobStatusViewSet.get_object', get_conversion_job)
+    def test_conversion_progress_when_created_is_new(self, *args, **kwargs):
+        conversion_job = get_conversion_job()
+
+        self.assertEqual(
+            min(conversion_job.gis_formats.values_list('progress', flat=True)),
```
das-g: When reading this test again, it strikes me as strange that we are comparing the minimum with an expected value. I think we should:
* compare the progress value **of each of the job's formats** with the expected value<br />or
* compare the `conversion_job.progress` **computed property** with the expected value (which will compute the minimum, but then that isn't the test's concern anymore)</br />or
* both

tests/conversion_job/view_test.py
das-g: This test shows that `_update_status_from_rq()` modifies the model, not the view. I think it should be a public method of the model, instead of a private method of the view.

PR #41, opened 2015-11-11
Feature/32 use locally available planet file

### LINKED ISSUE #32: Use locally available planet file
osmaxx-postgis-conversion should not care about how to obtain the planet file. Instead it should just require it to be at a predetermined local directory.
- [x] remove osmosis
- [x] use locally available planet file (to be provided by the deployer)

hixi:
Closes #32


tests/converters/boundaries_test.py
```diff
@@ -20,3 +22,27 @@ def test_init_when_parameters_are_missing_raises_type_error(self):
     def test_initializing_when_all_given_parameters_are_set_works(self):
         # shouldn't raise an error
         BBox(west=0, south=0, east=0, north=0)
+
+    @patch('converters.boundaries.BBox._get_cut_command', return_value='true')
+    def test_cut_pbf_calls_correctly(self, cut_command_mock):
```
das-g: `'/bin/true'` instead of just `'true'` would make clear the return value still is a command string and avoid confusion with other 'true' values, e.g. pythons `True` literal.

But instead of having the method under test call a known-to-succeed dummy command (`true`), how about checking whether it actually calls the command returned from the `_get_cut_command` call? As long as we assume that `subprocess.*` (and not `os.system` or `os.spawn*`) are being used, this could be probably be achieved by mocking `subprocess.Popen`. I guess all ways of executing a command with the `subprocess` module eventually instantiate that.

tests/converters/boundaries_test.py
```diff
@@ -20,3 +22,27 @@ def test_init_when_parameters_are_missing_raises_type_error(self):
...
+        bbox.cut_pbf(output_filename)
+        cut_command_mock.assertCalledWith(
+            output_filename=output_filename
+        )
+
+    @patch.dict('converters.converter_settings.OSMAXX_CONVERSION_SERVICE', {
+        'PBF_PLANET_FILE_PATH': '/path/to/planet-latest.osm.pbf',
+    })
+    # FIXME: _get_cut_command should also be sensitive to overriding django.conf.settings.OSMAXX_CONVERSION_SERVICE
+    #        but it isn't.
```
das-g: Is this something that can be fixed (see https://github.com/geometalab/osmaxx-postgis-conversion/pull/43#discussion-diff-44868302) or should I just remove `FIXME` comment?

PR #42, opened 2015-11-12
Feature/osmaxx 48 calculate estimated file size of excerpt data

hixi:
Integration of the branch from lukas.

Provides an api access under `/api/estimate_size_in_bytes/` which accepts post only requests and returns the given borders with corresponding size in bytes.

Closes geometalab/osmaxx#48


conversion_service/file_size_estimation/estimate_size.py
```diff
@@ -0,0 +1,101 @@
...
+
+def _estimate_size(west, south, east, north, size_dict):
+    # increase bbox by rounding outwards to the grid size
+    minlongrid = math.floor(west)
+    minlatgrid = math.floor(south)
+    maxlongrid = math.ceil(east)
+    maxlatgrid = math.ceil(north)
+
+    # main loop: iterate over all tiles that are partially or fully covered by the bbox
+    estimated_size = int(0)
```
das-g: `int(0)` is the same as `0`. But shouldn't this be a float, anyway? In what unit is this number, anyway?

conversion_service/file_size_estimation/estimate_size.py
```diff
@@ -0,0 +1,101 @@
...
+    for mylat in range(int(minlatgrid), int(maxlatgrid)):
+        for mylon in range(int(minlongrid), int(maxlongrid)):
+            ratio_value = _ratio(
+                max(mylon, west),
+                max(mylat, south),
+                min(mylon + 1, east),
+                min(mylat + 1, north)
+            )
```
das-g: Nice handling of precomputed data cells partially covered by the requested area.

Took me some time to figure out what you are computing here, though.

conversion_service/file_size_estimation/estimate_size.py
```diff
@@ -0,0 +1,101 @@
...
+            size_value = size_dict[str(mylat), str(mylon)]
+            estimated_size += size_value * ratio_value
```
das-g: How about using `sum( ... )` with a generator expression as argument? That should be more efficient than summing in a (nested) loop. (Don't use `sum( ... )` with a list comprehension, though!)

conversion_service/converters/converter_settings.py
```diff
@@ -1,9 +1,12 @@
+import os
+
 from django.conf import settings
 
 OSMAXX_CONVERSION_SERVICE = {
     'PBF_PLANET_FILE_PATH': '/var/data/osm-planet/planet-latest.osm.pbf',
     'RESULT_DIR': '/tmp/osmaxx-conversion-service/results',
     'RESULT_TTL': -1,  # never expire!
+    'ESTIMATION_CSV_SOURCE_FILE': os.path.join(os.path.dirname(__file__), '..', 'file_size_estimation', 'planet-stats.csv'),
```
wasabideveloper: Why do we not re use the path from `conversion_service/config/settings/common.py`?

conversion_service/file_size_estimation/estimate_size.py
hixi: I don't know, this code is mostly coming from the `lukas` branch on osmaxx, I tried to change as little as possible.

conversion_service/converters/converter_settings.py
hixi: We're using the same pattern as [django-rest-framework](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/settings.py), just much more simplified. 
This application is to be treated as a 'standalone' app as much as possible -> the settings in here are the 'default', and any setting defined in the application settings[1] will override this, if set. 

[1]: which happens to be `config/common.py` for us

conversion_service/file_size_estimation/estimate_size.py
hixi: This code is mostly coming from the `lukas` branch on osmaxx, I would like to avoid changing this just for the sake of speed - if by this change clarity could be improved, I think it would be worth the change though.

conversion_service/file_size_estimation/estimate_size.py
das-g: @ltog Can you answer this?

ltog: It has been a while since I wrote this... The csv file holds the size of each bbox file in bytes. It is calculated here: https://github.com/geometalab/osmaxx/blob/lukas/src/generate_stats.sh#L27

To estimate the download size for the end-user e.g. downloading a shapefile, this size will need to be scaled with a factor (which has to determined by experiment). The size in bytes can directly only be used as a measure how densely one grid is populated in comparison to another.

I hope that helps. Otherwise don't hestitate to ask further.

conversion_service/converters/converter_settings.py
wasabideveloper: :+1:

hixi:
@das-g Can we get this merged?

PR #43, opened 2015-11-14
Make the _get_cut_command test easier to understand than the code it's testing

das-g:


conversion_service/converters/boundaries.py
```diff
@@ -1,6 +1,6 @@
 import subprocess
 
-from converters.converter_settings import OSMAXX_CONVERSION_SERVICE
+from django.conf import settings
```
hixi: No, please don't change this line, you might want to write `from converters import converter_settings` but as it stands, this is wrong (you might not have any 'OSMAXX_CONVERSION_SERVICE' defined in your settings!).

tests/converters/boundaries_test.py
```diff
@@ -28,24 +26,21 @@ def test_initializing_when_all_given_parameters_are_set_works(self):
     @patch('converters.boundaries.BBox._get_cut_command', return_value='true')
     def test_cut_pbf_calls_correctly(self, cut_command_mock):
         # tests are using sample data from monaco
-        bbox = BBox(**BOUNDING_BOX_TEST_OSM)
+        bbox = BBox(west=1.23, south=-4.56, east=7.89, north=0.12)
```
hixi: Please use the Monaco coordinates (the ones from `BOUNDING_BOX_TEST_OSM`), otherwise the pbf file will always be empty (though we probably should write a test that covers `resultfile has content`).

tests/converters/boundaries_test.py
```diff
@@ -28,24 +26,21 @@ def test_initializing_when_all_given_parameters_are_set_works(self):
...
-            north=BOUNDING_BOX_TEST_OSM['north'],
-        )
-        bbox = BBox(**BOUNDING_BOX_TEST_OSM)
-        actual = bbox._get_cut_command(output_filename=output_filename)
+    @patch.dict('django.conf.settings.OSMAXX_CONVERSION_SERVICE', {
+        'PBF_PLANET_FILE_PATH': '/path/to/planet-latest.osm.pbf',
+    })
+    def test_get_cut_command_returns_expected_command(self, *args, **kwargs):
+        expected = "osmconvert --out-pbf -o=outfile.pbf -b=1.23,-4.56,7.89,0.12 /path/to/planet-latest.osm.pbf"
+        bbox = BBox(west=1.23, south=-4.56, east=7.89, north=0.12)
```
hixi: see remark above.

tests/osm_test_helpers.py
```diff
@@ -3,12 +3,5 @@
 
 from tests.conftest import test_data_dir
 
-BOUNDING_BOX_TEST_OSM = {
```
hixi: You might want to consider just naming this properly, not removing it altogether.

conversion_service/converters/boundaries.py
das-g: I had the impression that with the unchanged line, `django.conf.settings.OSMAXX_CONVERSION_SERVICE['PBF_PLANET_FILE_PATH']` is ignored **even if** it exists. I now realize that this isn't _really_ the case &mdash; just overwriting `django.conf.settings.OSMAXX_CONVERSION_SERVICE['PBF_PLANET_FILE_PATH']` _after_ `conversion_service/converters/converter_settings.py` has been run doesn't have any effect, because the [`OSMAXX_CONVERSION_SERVICE.update(settings.OSMAXX_CONVERSION_SERVICE)`](https://github.com/geometalab/osmaxx-postgis-conversion/blob/953b20fba874bcd08dd1021a4800ff9894332e02/conversion_service/converters/converter_settings.py#L10) doesn't get re-executed.

Will revert 192bfae0c3e6258eb323066d2d6b8c1f0747e1e5.

tests/converters/boundaries_test.py
das-g: Except that your test (unless I'm completely mistaken) doesn't ever cut out anything by running `osmconvert`. Instead, it runs `/bin/true`.

Further, I think this test is *right* in not actually calling `osmconvert` and we should not change that. (Ideally, it would not even call `true`.) If we want to test the actual cutting, this should be in another, additional test.

tests/converters/boundaries_test.py
das-g: The method tested here doesn't even call an external command. It just builds a string for doing so.

tests/converters/boundaries_test.py
hixi: You're right about that, I'm just mocking it (agree that it shouldn't test that it is cutting anything). So forget my remark, sorry for the noise.

PR #44, opened 2015-11-16
Feature/after work for pull request #40

hixi:
test amelioration for pull request #40 

To be reviewed by:
- [x] @das-g
- [X] @wasabideveloper

### DIFF
--- a/conversion_service/conversion_job/models.py
+++ b/conversion_service/conversion_job/models.py
@@ -1,5 +1,6 @@
 import os
 
+import django_rq
 from django.contrib.gis.db import models
 from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
@@ -8,7 +9,7 @@ from rest_framework.reverse import reverse
 from converters import CONVERTER_CHOICES, converter_settings
 from converters.boundaries import BBox
 from converters.converter import Options
-from shared import JobStatus, ConversionProgress
+from shared import JobStatus, ConversionProgress, rq_job_status_mapping
 from utils.directory_helper import get_file_only_path_list_in_directory
 
 
@@ -98,6 +99,21 @@ class ConversionJob(models.Model):
 
     def get_conversion_options(self):
         return Options(output_formats=self.gis_formats.values_list('format', flat=True))
+
+    def update_status_from_rq(self):
+        rq_job = django_rq.get_queue().fetch_job(job_id=self.rq_job_id)
+
+        # only do work if the job is not yet deleted
+        if rq_job:
+            self.status = rq_job_status_mapping[rq_job.status].value
+
+            progress = rq_job.meta.get('progress', None)
+            if progress:
+                progress_state = progress.value
+                for gis_format in self.gis_formats.all():
+                    gis_format.progress = progress_state
+                    gis_format.save()
+            self.save()
 
     @property
     def progress(self):
--- a/tests/conversion_job/model_test.py
+++ b/tests/conversion_job/model_test.py
@@ -8,6 +8,7 @@ from conversion_job.models import Extent, ConversionJob, GISFormat
 from converters import converter_settings, converter_options
 from converters.boundaries import BBox
 from shared import ConversionProgress
+from tests.conversion_job.view_test import django_rq_get_queue_stub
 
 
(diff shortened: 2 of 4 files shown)


tests/conversion_job/model_test.py
```diff
@@ -150,6 +151,26 @@ def test_progress_when_multiple_stati_are_set_returns_minimal_status_of_attached
...
+    @patch('django_rq.get_queue', django_rq_get_queue_stub)
+    def test_update_status_from_rq_sets_the_value_to_started(self, *args, **kwargs):
+        formats = converter_options.get_output_formats()
+        for out_format in formats:
+            GISFormat.objects.create(
+                conversion_job=self.conversion_job,
+                format=out_format
+            )
+
+        initial_progress_list = [ConversionProgress.NEW.value for _ in range(len(formats))]
```
das-g: Range construction is unnecessary here.

    initial_progress_list = [ConversionProgress.NEW.value for _ in formats]
or

    initial_progress_list = [ConversionProgress.NEW.value] * len(formats)
should give the same. Which do you prefer?

tests/conversion_job/model_test.py
```diff
@@ -150,6 +151,26 @@ def test_progress_when_multiple_stati_are_set_returns_minimal_status_of_attached
...
+        model_progress_list = list(self.conversion_job.gis_formats.values_list('progress', flat=True))
+        self.assertListEqual(model_progress_list, initial_progress_list)
+
+        started_progress_list = [ConversionProgress.STARTED.value for _ in range(len(formats))]
```
das-g: Another unnecessary `range` construction.

tests/conversion_job/view_test.py
```diff
@@ -47,21 +46,9 @@ def setUp(self):
...
-        conversion_job_status_view = ConversionJobStatusViewSet(kwargs={'rq_job_id': self.conversion_job.rq_job_id})
-        conversion_job_status_view._update_status_from_rq()
-
-        self.assertEqual(
-            min(conversion_job.gis_formats.values_list('progress', flat=True)),
-            ConversionProgress.STARTED.value
+        model_progress_list = list(conversion_job.gis_formats.values_list('progress', flat=True))
+        self.assertListEqual(
+            model_progress_list,
+            [ConversionProgress.NEW.value for _ in range(len(model_progress_list))]
```
das-g: Another unnecessary `range` construction.

conversion_service/conversion_job/models.py
```diff
@@ -99,6 +100,21 @@ def get_resulting_file_path_or_none(self, format):
     def get_conversion_options(self):
         return Options(output_formats=self.gis_formats.values_list('format', flat=True))
 
+    def update_status_from_rq(self):
```
wasabideveloper: Should the model really care about the queue? Maybe it's better to decouple this by a manager?

das-g: > Should the model really care about the queue?

Not sure about the queue, but this model should definitely care about (RQ) _jobs_, as we have this model to represent these jobs outside the (potentially volatile) queue. The bulk of this method's code is concerned with the relation between RQ job content and DB job model content, so the DB model seems like the right place for it. I like having this method here, because that allows to use it in a ["tell, don't ask"](https://pragprog.com/articles/tell-dont-ask) manner without too many levels of indirection.

> Maybe it's better to decouple this by a manager?

What would you move into the manager? The whole `update_status_from_rq()` method or the single line concerned with the queue?

hixi: let's stop over-engineering this; for now, the model instance is its own manager.

das-g:
> Some checks were not successful

It's cool that Travis is testing the code that would result from merging a pull request rather than only the to-be-pulled branch's code. But can we get it to also rebuild when the destination branch has new commits? That'd be even cooler!

The flake8 error [here](https://travis-ci.org/geometalab/osmaxx-postgis-conversion/jobs/91390090) was caused by a change that made it to `develop` and not to `feature/after_work_for_pull_request_#40`, but which has been fixed on `develop` by now.

hixi:
Unfortunately not, I already tried restarting it - and it still yields the same failure :-/

PR #45, opened 2015-11-16
Test private method _get_cut_command through public method cut_pbf

das-g:
... instead of directly testing the private method.

Also, don't make an actual call to an external command.


das-g:
Because `subprocess.Popen` is used as a context manager within `subprocess.call`, the prior would have been non-trivial to stub. Thus I've mocked the latter.

PR #48, opened 2015-11-20
Feature/cleaner choices enums

das-g:
Reviewed by:
- [x] @hixi

### DIFF
--- a/conversion_service/shared/__init__.py
+++ b/conversion_service/shared/__init__.py
@@ -36,13 +36,7 @@ class JobStatus(MostSignificantEnumMixin):
 
     @classmethod
     def choices(cls):
-        return (
-            (cls.NEW.value, 'new'),
-            (cls.QUEUED.value, 'queued'),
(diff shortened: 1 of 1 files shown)


conversion_service/shared/__init__.py
```diff
@@ -3,27 +3,27 @@
...
-                if status.precedence_over(most_significant_status):
-                    most_significant_status = status
-            return most_significant_status
+            return min(status_list, key=cls._precedence_list().index)
         else:
             return None
 
 
-class JobStatus(MostSignificantEnumMixin):
+class JobStatus(ChoicesEnum):
```
hixi: shouldn't this inherit from `MostSignificantEnumMixin` as well?

hixi: Ah, we don't use that feature for the `JobStatus`, since there is no overall status needed.

das-g: Ah, right. The _precedence_list is thus also not needed. Removing it in PR #49

PR #50, opened 2015-11-23
use explicit precedence values again

das-g:
also, distinguish between choice representation (on db and in html forms) and human readable choice name

reviewed by:
- [x] @hixi 
- [ ] @wasabideveloper

### DIFF
--- a/conversion_service/shared/__init__.py
+++ b/conversion_service/shared/__init__.py
@@ -10,14 +10,13 @@ class ChoicesEnum(enum.Enum):
 
 
 class MostSignificantEnumMixin(enum.Enum):
-    @staticmethod
-    def _precedence_list():
+    def precedence(self):
         raise NotImplementedError
 
     @classmethod
     def most_significant(cls, status_list):
         if len(status_list) > 0:
-            return min(status_list, key=cls._precedence_list().index)
+            return min(status_list, key=cls.precedence)
         else:
             return None
 
@@ -53,6 +52,9 @@ class ConversionProgress(ChoicesEnum, MostSignificantEnumMixin):
     def __init__(self, unique_name):
         self.technical_representation = unique_name
         self.human_readable_name = unique_name
+
+    def precedence(self):
+        return ConversionProgress._precedence_list().index(self)
 
     @staticmethod
     def _precedence_list():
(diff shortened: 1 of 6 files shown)


conversion_service/shared/__init__.py
```diff
@@ -29,6 +28,10 @@ class JobStatus(ChoicesEnum):
     STARTED = 'started'
     DONE = 'done'
 
+    def __init__(self, unique_name):
+        self.technical_representation = unique_name
+        self.human_readable_name = unique_name
```
hixi: I understand that technically, this is the same, but what if we wanted to keep the values the same but changing the human readable version? Or would you rather, in such a hypothetical case, make a migration that would update the database values as well?

conversion_service/shared/__init__.py
```diff
@@ -39,13 +42,19 @@ class JobStatus(ChoicesEnum):
...
+    STARTED = 'started', 2
+    SUCCESSFUL = 'successful', 3
+
+    def __init__(self, unique_name, precedence):
+        self.technical_representation = unique_name
+        self.human_readable_name = unique_name
+        self._precedence = precedence
+
+        # Allow lookup by unique_name:
+        type(self)._value2member_map_[unique_name] = self
```
hixi: :+1: I like this way of supplying both the value and precedence :-D

conversion_service/shared/__init__.py
das-g: Changing the human readable name would necessarily mean a Django migration, because the output of `choices()` (which is used in the model field definitions) would change.

Decoupling `technical_representation` and `human_readable_name` could be done by adding an argument to `__init__()` or by computing/looking up a name within `__init__()` from the existing argument. If not all arguments of `__init__()` shall be part of the Enum members' repsective `.value`, we can set `self._value_` in `__init__()` explicitly.

conversion_service/shared/__init__.py
das-g: I wonder whether explicitly setting `self._value_ = unique_name` would be even better. Lookup by the name-precedence tuple won't ever be necessary, I guess &mdash; precedence is only needed during Enum member initialization.

das-g: > I wonder whether explicitly setting self._value_ = unique_name would be even better.

I got this to work (I think &mdash; tests pass again, at least) by overriding `__new__()`: 24585af435cdab09c49a2373fc1cc1babc453146
This is harder to understand and might cause unexpected side effects, so I think we better stick to augmenting `_value2member_map_`.

hixi: Yes, sticking to augmenting `_value2member_map_` is better, IMHO.

PR #51, opened 2015-11-24
Feature/deployment

hixi:
Adaptations for easier deployment and support for sentry logging.
Reviewed by:

- [x] @das-g


conversion_service/config/settings/production.py
```diff
@@ -31,3 +34,58 @@
...
+                'handlers': ['console'],
+                'propagate': False,
+            },
+        },
+    }
+
+    RAVEN_CONFIG = {
+        'dsn': SENTRY_DSN,
+        # If you are using git, you can also automatically configure the
+        # release based on the git info.
```
das-g: We are using git. Should we do as hinted in this comment?

hixi: Oh, that didn't work, so I removed it - but I should remove the comment as well.

conversion_service/conversion_job/migrations/0003_auto_20151120_1528.py
```diff
@@ -0,0 +1,24 @@
...
+            model_name='conversionjob',
+            name='status',
+            field=models.CharField(max_length=20, verbose_name='job status', default='new', choices=[('error', 'error'), ('new', 'new'), ('queued', 'queued'), ('started', 'started'), ('done', 'done')]),
+        ),
+        migrations.AlterField(
+            model_name='gisformat',
+            name='progress',
+            field=models.CharField(max_length=20, verbose_name='progress', default='new', choices=[('error', 'error'), ('new', 'new'), ('received', 'received'), ('started', 'started'), ('successful', 'successful')]),
+        ),
+    ]
```
das-g: Can we avoid this migration? As far as I can see, it only changes the choices order. Now that we use explicit precedence values again since #50, we could change the order back to how it was.

hixi: Sure.

docker-compose.yml
```diff
@@ -30,7 +30,7 @@ api:
     service: environmentvars
   build: .
   dockerfile: ./Dockerfile.rest_api
-  command: [honcho, -f, ./conversion_service/Procfile, start]
+  command: [honcho, -f, ./conversion_service/Procfile.dev, start]
```
das-g: In `osmaxx-postgis-conversion`, are Procfiles only used for development?

hixi: Yes.

conversion_service/conversion_job/migrations/0003_auto_20151120_1528.py
das-g: Just deleting the migration (99e1a45) won't do. It'll be re-created when we run `manage.py makemigrations` the next time if we don't change the choices order back to how it was.

conversion_service/config/settings/production.py
das-g: :white_check_mark:

conversion_service/conversion_job/migrations/0003_auto_20151120_1528.py
hixi: I haven't changed the order of the choices, so I don't really know what you want here for a solution. Migrations will be created, and in the end, we can just squash them before release.

das-g: >  I don't really know what you want here

#52

hixi: OK, removed the migration _now_ ;-)

PR #53, opened 2015-11-25
Feature/create access user in init

hixi:
if user environment var is set: creates an user with a password if none exists, or changes the password if it does exist.

Reviewed by:
- [x] @das-g


Dockerfile.worker
```diff
@@ -108,10 +108,10 @@ RUN sed --in-place 's/\$\(CXX\) $\(LIBS\) -shared -o \$@ \$</\$\(CXX\) -shared -
...
-COPY ./conversion_service/worker/01_activate_translit.sql $HOME/conversion_service/worker/
-COPY ./conversion_service/worker/00_make_db_utf8.sh $HOME/conversion_service/worker/
-WORKDIR $HOME/conversion_service/worker/
+RUN mkdir -p $HOME/conversion_service/worker $HOME/entrypoint
+COPY ./docker_entrypoint/worker $HOME/entrypoint
+
+WORKDIR $HOME/entrypoint/
 
 USER postgres
 RUN service postgresql start && \
```
das-g: use fully qualified calls (using the variable) instead of changing the WORKDIR above

PR #54, opened 2015-11-27
Feature/fix url reversing error

hixi:
Fixes the url reversing issue in the worker.

Reviewed by:

- [x] @das-g

### DIFF
--- a/tests/worker/worker_test.py
+++ b/tests/worker/worker_test.py
@@ -39,17 +39,13 @@ class WorkerTest(TestCase):
         self.assertDictEqual(job_fetched.meta, {'progress': ConversionProgress.SUCCESSFUL})
 
 
-def do_nothing_callable():
-    pass
-
-
 class ConvertTest(TestCase):
     pbf_file_path = '/some_path/to/pbf.pbf'
     mocked_job_id = '0' * 36
     rq_job_stub = namedtuple('RQJob', ['id', 'meta', 'save'])(
         id=mocked_job_id,
         meta={'progress': ''},
-        save=do_nothing_callable,
+        save=mock.Mock(),
     )
 
     @mock.patch('os.remove')
--- a/tests/worker/worker_test.py
+++ b/tests/worker/worker_test.py
@@ -1,8 +1,10 @@
 # pylint: disable=C0111
+from collections import namedtuple
 from unittest import mock
 
 from django.test import TestCase
 from django_rq import get_connection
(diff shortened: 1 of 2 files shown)


tests/worker/worker_test.py
```diff
@@ -35,3 +37,33 @@ def test_set_progress_on_job_sets_the_progress(self):
...
+    pass
+
+
+class ConvertTest(TestCase):
+    pbf_file_path = '/some_path/to/pbf.pbf'
+    mocked_job_id = '0' * 36
+    rq_job_stub = namedtuple('RQJob', ['id', 'meta', 'save'])(
+        id=mocked_job_id,
+        meta={'progress': ''},
+        save=do_nothing_callable,
```
das-g: Why not use a `unittest.mock.Mock` object here? (Without `@patch`). One of their purposes is to be callable objects.

tests/worker/worker_test.py
```diff
@@ -35,3 +37,29 @@ def test_set_progress_on_job_sets_the_progress(self):
...
+    @mock.patch('converters.gis_converter.bootstrap.bootstrap.boostrap', return_value=None)
+    @mock.patch('converters.osm_cutter.cut_osm_extent', return_value=pbf_file_path)
+    @mock.patch('rq.get_current_job', return_value=rq_job_stub)
+    @mock.patch('worker.converter_job.Notifier')
+    def test_convert_with_host_and_job_assembles_url_correctly(self, notifier_mock, rq_job_stub, *args, **kwargs):
+        format_options = Options(output_formats=['fgdb', 'spatialite', 'shp', 'gpkg'])
+        geometry, output_directory, callback_url = None, '/tmp/', None
+        host = 'http://converter-host.example.com'
+        convert(geometry, format_options, output_directory, callback_url, host)
+        expected_url = host + reverse(viewname='gisformat-detail', kwargs={'pk': self.mocked_job_id})
```
das-g: Can we hardcode the `expected_url` value instead of computing it the same way the implementation does? Just repeating the implementation in the test doesn't give much confidence in its correctness.

hixi: what is not giving confidence? we're using the reverse from the 3rd party library, and if we ever change the regex, this test will adapt correctly. Even though It is the same code, it does test the right thing IMHO.

das-g: Convinced.

:white_check_mark: leave as-is

PR #56, opened 2015-11-27
assert model is saved; create model and then save again

hixi:
Fixes issue with path unavailable.

reviewed by:
 - [x] @das-g 
 - [x] @hixi

### DIFF
--- a/conversion_service/conversion_job/models.py
+++ b/conversion_service/conversion_job/models.py
@@ -77,6 +77,8 @@ class ConversionJob(models.Model):
 
     @property
     def output_directory(self):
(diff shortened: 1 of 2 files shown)


conversion_service/conversion_job/models.py
```diff
@@ -77,6 +77,10 @@ class ConversionJob(models.Model):
 
     @property
     def output_directory(self):
+        """
+        Only available on saved model instances.
+        """
```
das-g: changed comment to Docstring, please review

hixi: :+1:

PR #58, opened 2015-12-04
Add workaround for strange docker permission issue

das-g:
To get rid of `Starting PostgreSQL 9.3 database server: mainThe PostgreSQL server failed to start. Please check the log output: [...] [23-1] FATAL: could not access private key file "/etc/ssl/private/ssl-cert-snakeoil.key": Permission denied ... failed!` on das-g's system.

See [j0hnsmith's comment docker/docker#783](https://github.com/docker/docker/issues/783\#issuecomment-56013588).

###### Reviewed by:
- [x] @wasabideveloper
- [x] @hixi

### DIFF
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -107,6 +107,10 @@ RUN sed --in-place 's/\$\(CXX\) $\(LIBS\) -shared -o \$@ \$</\$\(CXX\) -shared -
 
 ENV HOME /home/py
 
+# Workaround for AUFS-related(?) permission issue:
+# See https://github.com/docker/docker/issues/783#issuecomment-56013588
+RUN mkdir /etc/ssl/private-copy; mv /etc/ssl/private/* /etc/ssl/private-copy/; rm -r /etc/ssl/private; mv /etc/ssl/private-copy /etc/ssl/private; chmod -R 0700 /etc/ssl/private; chown -R postgres /etc/ssl/private
+
 # activate translit
 RUN mkdir -p $HOME/conversion_service/worker $HOME/entrypoint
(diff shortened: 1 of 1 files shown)


das-g:
We're not the only ones doing this b.t.w.: https://github.com/orchardup/docker-postgresql/blob/955ac7787440aca56d32906e7338fd4ca8d9cb27/Dockerfile#L11

I guess the [official postgres image](https://github.com/docker-library/postgres/blob/ed23320582f4ec5b0e5e35c99d98966dacbc6ed8/9.3/Dockerfile) doesn't need this, because it's Debian- and not Ubuntu-based.

hixi:
Would switching to debian:jessie be an issue (ie a lot of work)? If it really doesn't happen on debian, I would like that more than this workaround.

wasabideveloper:
Didn't we early use an debian image? Why did we switch to Ubuntu? I like Ubuntu on desktop but I don't have experiences how it performs on servers.

das-g:
Eh, I just realized that our worker image _is_ `debian:jessie`-based. (Via [`geometalab/gdal-debian`](https://hub.docker.com/r/geometalab/gdal-debian))

hixi:
That's what I thought, because I really tried to use debian-based images whenever/wherever possible, but I might have missed this one ;-)

Then I'm in favor of merging this, because it enables development again; before merging, could you @das-g create an issue so we can investigate this another time, I don't particularly like workarounds on such a low level.

das-g:
> create an issue so we can investigate this another time

Issue filed: geometalab/osmaxx#283

PR #59, opened 2015-12-04
Feature/osmaxx 166 finish administrative areas support

hixi:
Reviewed by:

- [x] @das-g 
- [x] @wasabideveloper

Is part of geometalab/osmaxx#166


conversion_service/converters/boundaries.py
```diff
@@ -32,6 +32,32 @@ def _get_cut_command(self, output_filename):
...
+        self.polyfile_path = country_polyfile_path
+
+    def cut_pbf(self, output_filename):
+        command = self._get_cut_command(output_filename=output_filename)
+        subprocess.check_call(command.split())
+        return output_filename
+
+    def _get_cut_command(self, output_filename):
+        pbf_file_path = OSMAXX_CONVERSION_SERVICE.get('PBF_PLANET_FILE_PATH')
+        return "osmconvert --out-pbf -o={output_filename} -B={polyfile_path} {pbf_file_path}".format(
```
wasabideveloper: What if the file name contains whitespaces?

conversion_service/countries/storage.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from django.conf import settings
+from django.core.files.storage import FileSystemStorage
+
+polyfile_location = settings.OSMAXX_CONVERSION_SERVICE.get('COUNTRIES_POLYFILE_LOCATION') \
+    or os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data', 'polyfiles')
```
wasabideveloper: Do we really want to store the files in a subdirectory of this code file if COUNTRIES_POLYFILE_LOCATION is undefined or do I missunderstand this line?

conversion_service/converters/boundaries.py
hixi: It will fail loudly.

See https://github.com/geometalab/osmaxx-postgis-conversion/blob/3bbadb50b769f047b21f6099ff824612cc89c076/tests/converters/boundaries_test.py#L41 (we know that, we just hope someone else is fixing the issue :-P)

conversion_service/conversion_job/models.py
```diff
@@ -18,13 +19,16 @@ class Extent(models.Model):
...
 
     def clean(self, exclude=None, validate_unique=True):
         if self._bbox_partially_present() and not self._bbox_present():
             raise ValidationError(_('incomplete bounding box boundaries'))
-        if not(self._bbox_present() ^ self._polyfile_present()):
-            raise ValidationError(_('either extents or polyfile must be given'))
+        if not(self._bbox_present() ^ self._country_present()):
+            raise ValidationError(_('either extents or country must be given'))
+        if self._polyfile_present():
+            raise ValidationError(_('polyfile is not supported (deprecated)'))
```
das-g: Would be friendly to list available alternatives with the deprecation message. But as this is an internal interface, we can probably handle without that extra information.

conversion_service/countries/storage.py
hixi: This is not actual storage, but the location (for the import during the data migration 0002).

conversion_service/conversion_job/models.py
hixi: especially since polyfile _never ever_ worked :smirk:

conversion_service/countries/storage.py
wasabideveloper: Please add this reason as comment if you think it's necessary.

:heavy_check_mark:

conversion_service/countries/utils.py
```diff
@@ -0,0 +1,70 @@
...
+from django.contrib.gis.geos import MultiPolygon, Polygon, GEOSGeometry
+
+from countries.storage import polyfile_location
+
+
+def get_polyfile_name_to_file_mapping():
+    polyfile_mapping = {}
+    for possible_polyfile in os.listdir(polyfile_location):
+        if possible_polyfile.endswith('.poly'):
+            name = possible_polyfile.split('.poly')[0]
```
das-g: Should `'.poly'` be a constant?

And what if a filename doesn't end in `'.poly'`? Shouldn't that at least cause a warning that the full name will be used?

PR #61, opened 2015-12-08
Feature/fix callback issues

hixi:
Reviewed by:

A `bug fix` which fixes delivering the wrong callback url.

- [x] @das-g 

```yoda
- more tests add we should
- through another issue they covered will be
```

See geometalab/osmaxx#291

### DIFF
--- a/conversion_service/worker/converter_job.py
+++ b/conversion_service/worker/converter_job.py
@@ -55,12 +55,12 @@ class Notifier(object):
         if not self.noop:  # pragma: nocover
             data = {'status': self.status_url} if self.status_url is not None else {}
             try:
-                requests.get(self.callback_url, data=data)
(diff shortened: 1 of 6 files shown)


conversion_service/worker/converter_job.py
```diff
@@ -55,12 +55,12 @@ def _notify_status_change(self):
...
             data = {'status': self.status_url} if self.status_url is not None else {}
             try:
-                requests.get(self.callback_url, data=data)
+                requests.get(self.callback_url, params=data)
             except:
                 pass
 
 
-def convert(geometry, format_options, output_directory, callback_url, host=None):
+def convert(geometry, format_options, output_directory, callback_url, protocol=None, host=None):
```
das-g: `None` is only a good default for `protocol` if it's treated specially.

PR #63, opened 2015-12-10
Feature/garmin

hixi:
Adds Garmin support.

Reviewed by:
- [x] @das-g
- [x] @wasabideveloper

Part of geometalab/osmaxx#154


conversion_service/converters/converter.py
```diff
@@ -1,9 +1,82 @@
...
+from utils import chg_dir_with
 
-    def __add__(self, other):
-        return Options(output_formats=self.output_formats + other.output_formats)
+
+class Conversion(object):
+    def __init__(self, formats, output_dir, osm_pbf_path, basename='osmaxx_excerpt'):
+        self.formats = formats
+        self.output_dir = output_dir
+        self.filename_prefix = '_'.join([
```
wasabideveloper: Maybe it makes sense to move this to configuration.

conversion_service/converters/converter.py
```diff
@@ -1,9 +1,82 @@
...
+import shutil
+import subprocess
+import time
 
-    def get_output_formats(self):
-        return self.output_formats
+from converters import garmin_converter
+from converters import gis_converter
+from converters.gis_converter.bootstrap import bootstrap
+from utils import chg_dir_with
```
wasabideveloper: What's this function doing? Maybe it's more readable using a understandable alias on import.

conversion_service/converters/converter.py
```diff
@@ -1,9 +1,82 @@
...
+        self._extract_postgis_formats(gis_formats)
+
+    def _extract_postgis_formats(self, formats):
+        if len(formats) > 0:
+            bootstrap.boostrap(self.pbf_path)
+            with chg_dir_with(os.path.dirname(__file__)):
+                # only create statistics once and remove it when done with all formats
+                self._get_statistics(self.tmp_statistics_filename)
+                for format in formats:
+                    file_basename = '_'.join([self.filename_prefix, format])
```
wasabideveloper: Maybe it makes sense to move this to configuration.

conversion_service/converters/converter.py
```diff
@@ -1,9 +1,82 @@
...
+                    self._copy_statistics_file_to_format_dir(file_basename)
+                    self._export_from_db_to_format(file_basename, format)
+                # remove the temporary statistics file
+                os.remove(os.path.join(self.output_dir, 'tmp', self.tmp_statistics_filename + '_STATISTICS.csv'))
```
wasabideveloper: Maybe it's better to define a CONSTANT and user everywhere needed.

conversion_service/converters/converter.py
```diff
@@ -1,9 +1,82 @@
...
+
+    # Export files of the specified format (file_format) from existing database
+    def _export_from_db_to_format(self, file_basename, file_format):
+        extract_base_dir = os.path.join(os.path.dirname(__file__), 'gis_converter', 'extract')
+        extract_format_file_path = os.path.join(extract_base_dir, 'extract', 'extract_format.sh')
+        extra_data_dir = os.path.join(extract_base_dir, 'static')
+
+        dbcmd = 'sh', extract_format_file_path, self.output_dir, file_basename, file_format, extra_data_dir
+        dbcmd = [str(arg) for arg in dbcmd]
+        subprocess.check_call(dbcmd)
```
wasabideveloper: Maybe we can merge this and the previous or the previous and second last line because the lines are as short.

conversion_service/converters/converter.py
```diff
@@ -1,9 +1,82 @@
...
+
+    # Extract Statistics
+    def _get_statistics(self, filename):
+        extract_statistics_file_path = os.path.join(
+            os.path.dirname(__file__), 'gis_converter', 'extract', 'extract', 'extract_statistics.sh'
+        )
+        statcmd = 'bash', extract_statistics_file_path, self.output_dir, filename
```
wasabideveloper: If we use shortcuts I think it's more readable to use snake case: `stat_cmd`. Better would be `statistic_command`.

conversion_service/converters/converter.py
```diff
@@ -1,9 +1,82 @@
...
+        statcmd = [str(arg) for arg in statcmd]
+        subprocess.check_call(statcmd)
+
+    def _copy_statistics_file_to_format_dir(self, file_basename):  # pragma: nocover
+        shutil.copyfile(
+            os.path.join(self.output_dir, 'tmp', self.tmp_statistics_filename + '_STATISTICS.csv'),
```
wasabideveloper: See comment above about CONSTANTS.

conversion_service/converters/converter.py
```diff
@@ -1,9 +1,87 @@
...
+                # only create statistics once and remove it when done with all formats
+                self._get_statistics(self.tmp_statistics_filename)
+                for format in formats:
+                    file_basename = '_'.join([self.filename_prefix, format])
+                    self._copy_statistics_file_to_format_dir(file_basename)
+                    self._export_from_db_to_format(file_basename, format)
+                # remove the temporary statistics file
+                os.remove(os.path.join(self.output_dir, 'tmp', self.tmp_statistics_filename + '_STATISTICS.csv'))
+
+    def _create_garmin_export(self, formats):
```
das-g: Would it make sense to `assert len(formats) <= 1` here?

conversion_service/converters/converter.py
hixi: it does what it says: changes directory usable in a `with` statement.

conversion_service/converters/converter.py
hixi: No, then we would always have the same time-stamp, which doesn't make sense.

conversion_service/converters/converter.py
hixi: `assert len(formats) == 1` can be asserted ;-)

conversion_service/converters/converter.py
wasabideveloper: I mean the structure of the filename prefix. E.q 
```python
FILENAME_PREFIX = '{basename}_{time}'
```

conversion_service/converters/converter.py
das-g: Maybe it could be renamed to `changed_dir(...)`, so that usage would read
```python
with changed_dir('/my/directory') as workingdir:
    # do something in workingdir ...
    # ...

# You're back to previous working dir once you leave the `with` clause
# ...
```

conversion_service/converters/converter.py
hixi: oh, scratch that; this check is to make sure we even want to do a garmin extraction, since it's can be called with an empty list or None.

conversion_service/converters/converter.py
hixi: I don't think this would make much sense given the rather dynamic nature of this file naming.

conversion_service/converters/converter.py
hixi: A constant isn't the right thing here, since it will have a dynamic part; but extracting this into a variable/function/calculated property does make sense.

conversion_service/converters/converter.py
das-g: I'd keep that as-is. I think it's more readable than when this is inlined.

conversion_service/converters/converter.py
das-g: Agreed.

conversion_service/converters/converter.py
hixi: this is not a constant ;-)

conversion_service/converters/converter.py
das-g: But we still expect 0 or 1 formats here, no more, right? So `<=` would make sense, wouldn't it?

conversion_service/converters/converter.py
wasabideveloper: How about this ;-)
```python
self.filename_prefix = settings.osmaxx['FILENAME_PREFIX'].format(
    basename=basename,
    time=time.strftime("%Y-%m-%d_%H%M%S"),
)
```
The reason: I think everywhere we create file or directory names or parts we should move the **name structure** to configuration. I did this for the download file name and weeks later we changed the download file again and again. I was so happy I didn't had to search the code if there is some place I need to change it too ;-).

hixi: This is internal, and the file will be renamed when the client (osmaxx) downloads it. I dislike settings for no reasons, we already have very many of them - if we _really_ need it later, it is really easy to add it.

PR #67, opened 2015-12-18
remove countries with too large polyfile from db

hixi:
reviewed by:
- [x] @das-g 
- [ ] @wasabideveloper 

kept the polyfiles themselves around, so backwards migration is possible.


das-g:
:+1:

PR #69, opened 2015-12-22
worker should be allowed to work for 24 hours

hixi:
...without timing out.

A queued job shold not timeout for a day.

Reviewed by:

- [x] @das-g 
- [x] @wasabideveloper

### DIFF
--- a/conversion_service/manager/rq_helper.py
+++ b/conversion_service/manager/rq_helper.py
@@ -10,7 +10,7 @@ def rq_enqueue_with_settings(function, *args, **kwargs):
         function,
         result_ttl=converter_settings.OSMAXX_CONVERSION_SERVICE['RESULT_TTL'],
         ttl=timedelta(days=1).seconds,  # max time in the queue, before it is considered lost
-        timeout=timedelta(hours=10).seconds,  # max time the worker may spend, before the job is cancelled
+        timeout=timedelta(days=1).seconds,  # max time the worker may spend, before the job is cancelled
         *args,
         **kwargs
     )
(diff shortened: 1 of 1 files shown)


conversion_service/manager/rq_helper.py
```diff
@@ -1,12 +1,16 @@
...
-            function,
-            result_ttl=converter_settings.OSMAXX_CONVERSION_SERVICE['RESULT_TTL'],
-            *args,
-            **kwargs
-        )
+    return django_rq.enqueue(
+        function,
+        result_ttl=converter_settings.OSMAXX_CONVERSION_SERVICE['RESULT_TTL'],
+        ttl=timedelta(days=1).seconds,  # max time in the queue, before it is considered lost
+        timeout=timedelta(hours=10).seconds,  # max time the worker may spend, before the job is cancelled
```
hixi: garmin extraction is quite fast: about 20-30 minutes for Switzerland. The so-called `gis`-conversions not so much: about 2h (and something) for Switzerland alone.

I guess we should either increase this value further (up to, lets say, 2 days?), so that even big, high dense mapped countries finish or drop the support for those countries and only allow smaller sub-parts (ie. Germany: only *Bundesländer*, USA: only *States*).

I think the best results can be achieved by doing both:
- Big countries with a low mapping percentage will grow, and we still want it to succeed
- Where we already know it is `too big to extract` :tm:, we only support sub-regions

hixi:
As discussed, 24h *should* be enough for big countries to finish as well.

PR #73, opened 2015-12-30
Feature/better sql testing support

hixi:
Part of geometalab/osmaxx#334

* refactoring to enable easier testing (later on).
* removed all shell scripts and replaced with python equivalents

Bug fix:
* set the DEFAULT_TIMEOUT to one day, which seems to be the upper limit of a queue, override can only lower that number (though this is undocumented in those libraries).

_Accidentally_ closes geometalab/osmaxx#354

Reviewed by:

- [x] @das-g 

Please be aware, that there are no tests; if the direction is good, I'll add them at a later point.


conversion_service/converters/gis_converter/extract/statistics/statistics.py
```diff
@@ -0,0 +1,261 @@
+from converters.gis_converter.extract.statistics.osm_groups import BOUNDARIES, SETTLEMENTS, LANDUSES, MILITARY_AREAS, \
+    MISC_LINES, NATURAL_AREAS, NATURAL_POIS, HIGHWAY_RAILWAY_TYPES, PLACES_OF_WORSHIP, PUBLIC_POIS, RECYCLING_POIS, \
+    EDUCTATION_POIS, HEALTH_POIS, LEISURE_POIS, CATERING_POIS, ACCOMODATION_IN_POIS, ACCOMODATION_OUT_POIS, SHOP_POIS, \
+    VENDING_POIS, MONEY_POIS, TOURISM_POIS, DESTINATION_POIS, MISC_POIS, SPORT, MAN_MADE, HISTORIC, AMENITY, \
+    RAILWAY_TYPES, ROAD_TYPES, ROUNDABOUT_TYPES, ROUTE_TYPES, TRAFFIC_AREA_TYPES, TRAFFIC_POINTS_TYPES, \
+    TRANSPORT_AREAS_TYPES, UTILITY_AREA_TYPES, UTILITY_POINTS_TYPES, UTILITY_LINES_TYPES, PIPELINE, WATER_AREA_TYPES, \
+    WATER_POINT_TYPES, WATER_LINE_TYPES
```
das-g: How about
```python
from converters.gis_converter.extract.statistics.osm_groups import *
```
, here?

hixi: might be a good idea, even though `*` imports are usually not good - this might just be the exception to confirm that rule ;-)

conversion_service/converters/converter_settings.py
```diff
@@ -11,3 +11,10 @@
 
 if hasattr(settings, 'OSMAXX_CONVERSION_SERVICE'):
     OSMAXX_CONVERSION_SERVICE.update(settings.OSMAXX_CONVERSION_SERVICE)
+
+# internal values, can't be overridden but are kept centralized
```
das-g: What does "can't be overridden" mean here?

- Can they be changed by editing the following lines?
- Must something be kept in sync if the following lines are edited? If so, what?
- Are calls to `OSMAXX_CONVERSION_SERVICE.update()` elsewhere than here (e.g. in test `setUp()` methods) without effect?

hixi: They do not follow the same principle as the other configuration values, that you can override in the config/local.py etc. If these lines are changed, the sql scripts most likely won't work anymore, but they _might_.
They just moved from (AFAIK) 3 places to here, so they are not hard-coded in the scripts anymore.

conversion_service/converters/garmin_converter/to_garmin.py
```diff
@@ -57,9 +61,9 @@ def _produce_garmin(self, config_file_path, workdir, out_dir):
             config
         )
 
-    def _zip_and_move(self, workdir):
+    def _create_zip(self, workdir):
```
das-g: Argument `workdir` should probably be renamed. Not quite sure to what, though.

hixi: why? what's the issue yout seeing with `workdir`?

conversion_service/converters/converter_settings.py
das-g: :white_check_mark: less confusing with da8d2b4

& thanks for the explanation &mdash; this is good to know.

conversion_service/converters/garmin_converter/to_garmin.py
das-g: `workdir` reminds of the UNIX concept of 'current working dir'. This argument isn't nessesarily the current working dir. (It might be, but it shouldn't matter as far as I can see.)

Maybe just call it `dir`?

hixi: `dir` is a [keyword](https://docs.python.org/3/library/functions.html#dir) in python.

conversion_service/converters/gis_converter/bootstrap/bootstrap.py
```diff
@@ -1,10 +1,100 @@
...
+            '--tag-transform-script', self._style_path,
+            '--number-processes', '8',
+            '--username', postgres_user,
+            '--hstore-all',
+            '--input-reader', 'pbf',
+            self._pbf_file_path,
+        ]
+
+        if self._limit_ram_usage:
+            osm_2_pgsql_command.insert(2, '--slim')
```
das-g: Magic number `2`. Also, this looks somewhat fragile. Should `'--slim'` be inserted after `'--create'` or before `'--extra-attributes'`? (Same effect currently, but `osm_2_pgsql_command` could change.)

Dunno if something like this would be better:
```python
        if self._limit_ram_usage:
             insert_pos = osm_2_pgsql_command.index('--extra-attributes')
             osm_2_pgsql_command.insert(insert_pos, '--slim') 
```

hixi: I just positioned this at the same place as before, I like the explicitness of your proposition, but for me personally, I find the magic number intuitive enough.

conversion_service/converters/gis_converter/bootstrap/bootstrap.py
```diff
@@ -1,10 +1,100 @@
...
+
+        subprocess.check_call(osm_2_pgsql_command)
+
+    def _setup_db_functions(self):
+        create_function_sql_path = os.path.join(self._script_base_dir, 'sql', 'create_functions.sql')
+        sql = open(create_function_sql_path, "r").read()
```
das-g: The file descriptor will be closed when the return value of `open()` is garbage-collected. This can be immediately after this line (as the fd isn't being assigned to anything, but only used to call `read()` on it) or any time later.

To force immediate closing, consider using a `with` statement.

conversion_service/converters/gis_converter/bootstrap/bootstrap.py
```diff
@@ -1,10 +1,100 @@
...
+        self._postgres.execute_psql(sql)
```
das-g: Why not use `sql_file_path=` instead of reading the script into python's memory? That way, we'd never hold a file descriptor.

conversion_service/converters/gis_converter/bootstrap/bootstrap.py
```diff
@@ -1,10 +1,100 @@
...
+
+    def _harmonize_database(self):
+        cleanup_sql_path = os.path.join(self._script_base_dir, 'sql', 'sweeping_data.sql')
+        sql = open(cleanup_sql_path, "r").read()
```
das-g: As above: The file descriptor will be closed when the return value of `open()` is garbage-collected. This can be immediately after this line (as the fd isn't being assigned to anything, but only used to call `read()` on it) or any time later.

To force immediate closing, consider using a `with` statement.

conversion_service/converters/gis_converter/bootstrap/bootstrap.py
```diff
@@ -1,10 +1,100 @@
...
+        self._postgres.execute_psql(sql)
```
das-g: As above: Why not use `sql_file_path=` instead of reading the script into python's memory? That way, we'd never hold a file descriptor.

conversion_service/converters/gis_converter/helper/postgres_wrapper.py
```diff
@@ -0,0 +1,82 @@
...
+    def execute_raw(self, sql, connection=None, autocommit=False, fetch_result=True):
+        if connection is None:
+            connection = self._get_connection()
+        if autocommit:
+            connection.set_isolation_level(psycopg2.extensions.ISOLATION_LEVEL_AUTOCOMMIT)
+        cursor = connection.cursor()
+        cursor.execute(sql)
+        return cursor
+
+    def execute_psql(self, sql=None, sql_file_path=None, extra_args=None):
```
das-g: Consider making two public functions for this, with simpler signatures and contracts:
- `execute_psql_command(self, sql_string_or_single_psql_backslash_command, extra_args=None)`
- `execute_psql_script(self, psql_script_file_path, extra_args=None)`

conversion_service/converters/gis_converter/bootstrap/sql/filter/address.sql
```diff
@@ -130,6 +130,7 @@ select
   addr_interpolate(line_id, addr_street, interpolation_type, first_housenr, last_housenr, line_geom)
 from addr_interpolated
 limit 1 \g ./tmp/FORCE_CONTINUE.txt;
+-- last line: limit 1 \g ./tmp/FORCE_CONTINUE.txt; -> what does that do?
```
das-g: This `SELECT` might have a long-ish output, which might cause `psql` to display it using a paging tool (e.g. `more` or `less`) that would not return without user interaction (typing 'q' for the mentioned pagers). To not make the script dependent on user input, the output is redirected to some file.

Why the `SELECT` is made at all, I don't know. Maybe `addr_interpolate(...)` has a side effect? @sfkeller: Can you shed some light on this?

conversion_service/converters/gis_converter/bootstrap/sql/filter/drop_and_recreate.sql
```diff
@@ -31,6 +31,8 @@ DROP VIEW if exists view_osmaxx.water_p;
 DROP VIEW if exists view_osmaxx.water_a;
 DROP VIEW if exists view_osmaxx.water_l;
 
+DROP SCHEMA IF EXISTS view_osmaxx;
+
 -- DROP SCHEMA IF EXISTS osmaxx;
```
das-g: This commented-out code is now obsolete, I guess, as you drop the SCHEMA below?

conversion_service/converters/gis_converter/bootstrap/sql/filter/drop_and_recreate.sql
```diff
@@ -61,3 +63,9 @@ DROP TABLE if exists osmaxx.utility_l;
 DROP TABLE if exists osmaxx.water_p;
 DROP TABLE if exists osmaxx.water_a;
 DROP TABLE if exists osmaxx.water_l;
+
+DROP SCHEMA IF EXISTS osmaxx;
```
das-g: If you drop the SCHEMA, anyway, don't bother about dropping its TABLEs above.

conversion_service/converters/gis_converter/bootstrap/sql/filter/drop_and_recreate.sql
```diff
@@ -31,6 +31,8 @@ DROP VIEW if exists view_osmaxx.water_p;
 DROP VIEW if exists view_osmaxx.water_a;
 DROP VIEW if exists view_osmaxx.water_l;
 
+DROP SCHEMA IF EXISTS view_osmaxx;
```
das-g: If you drop the SCHEMA, anyway, don't bother about dropping its TABLEs above.

conversion_service/converters/gis_converter/extract/db_to_format/extract.py
```diff
@@ -0,0 +1,54 @@
...
+
+    to_format_options = FORMATS[to_format]
+    extraction_options = to_format_options['extraction_options']
+    ogr_name = to_format_options['ogr_name']
+    extension = to_format_options['extension']
+
+    output_path = os.path.join(output_dir, base_filename + extension)
+
+    ogr2ogr_command = [
+        'ogr2ogr', '-f', '{0}'.format(ogr_name), output_path,
```
das-g: `'{0}'.format(ogr_name)` should be the same as `str(ogr_name)` or (assuming it already is a string) simply `ogr_name`, shouldn't it?

conversion_service/converters/gis_converter/extract/statistics/statistics.py
das-g: Maybe in combination with a decent `__all__`, so that we don't get the `COMMON_...` constants?

conversion_service/converters/gis_converter/extract/statistics/osm_groups.py
```diff
@@ -0,0 +1,44 @@
...
+TRAFFIC_POINTS_TYPES = 'general_traffic traffic_signals mini_roundabout stop crossing level_crossing speed_camera motorway_junction turning_circle ford street_lamp barrier entrance gate bollard lift_gate stile cycle_barrier fence toll_booth block kissing_gate cattle_grid traffic_calming hump bump table chicane cushion fuel services parking surface multi-storey underground bicycle'.split()
+TRANSPORT_AREAS_TYPES = 'railway_station railway_halt bus_stop bus_station taxi_stand airport runway helipad ferry_terminal aerialway_station aeroway aerialway stop_position taxiway apron'.split()
+UTILITY_AREA_TYPES = 'tower station nuclear solar fossil hydro wind substation transformer water_works wastewater_plant power'.split()
+UTILITY_POINTS_TYPES = UTILITY_AREA_TYPES + 'pole'.split()
+UTILITY_LINES_TYPES = 'line minor_line cable minor_cable power'.split()
+PIPELINE = 'pipeline'.split()
+COMMON_WATER_TYPES = 'water spring riverbank slipway marina pier dam weir reservoir_covered waterway'.split()
+WATER_AREA_TYPES = COMMON_WATER_TYPES
+WATER_POINT_TYPES = COMMON_WATER_TYPES + 'waterfall lock_gate'.split()
+WATER_LINE_TYPES = 'river stream canal drain waterway'.split()
```
das-g: I wonder whether these should be sets instead of lists.

Also, should we put this into some YAML file or something, maybe?

conversion_service/converters/gis_converter/extract/statistics/statistics.py
```diff
@@ -0,0 +1,261 @@
...
+from converters.gis_converter.helper.default_postgres import get_default_postgres_wrapper
+
+
+def gather_statistics(outfile):
+    out_file = open(outfile, 'w')
+    all_stats = Statistics().gather_all()
+    for stat in all_stats:
+        print(';'.join(stat), file=out_file)
+    out_file.close()
```
das-g: consider using a `with`-statement

conversion_service/converters/gis_converter/extract/statistics/statistics.py
```diff
@@ -0,0 +1,261 @@
...
+    VENDING_POIS, MONEY_POIS, TOURISM_POIS, DESTINATION_POIS, MISC_POIS, SPORT, MAN_MADE, HISTORIC, AMENITY, \
+    RAILWAY_TYPES, ROAD_TYPES, ROUNDABOUT_TYPES, ROUTE_TYPES, TRAFFIC_AREA_TYPES, TRAFFIC_POINTS_TYPES, \
+    TRANSPORT_AREAS_TYPES, UTILITY_AREA_TYPES, UTILITY_POINTS_TYPES, UTILITY_LINES_TYPES, PIPELINE, WATER_AREA_TYPES, \
+    WATER_POINT_TYPES, WATER_LINE_TYPES
+from converters.gis_converter.helper.default_postgres import get_default_postgres_wrapper
+
+
+def gather_statistics(outfile):
+    out_file = open(outfile, 'w')
+    all_stats = Statistics().gather_all()
```
das-g: Does it make sense to have file open during the gathering already? (Maybe so we make sure we only gather statistics if we have somewhere to write them?)

conversion_service/converters/gis_converter/extract/statistics/statistics.py
```diff
@@ -0,0 +1,261 @@
...
+    for stat in all_stats:
+        print(';'.join(stat), file=out_file)
```
das-g: Consider using Pythons [CSV API](https://docs.python.org/3/library/csv.html).

conversion_service/converters/gis_converter/bootstrap/bootstrap.py
hixi: IIRC I did this after trying to get some useful information out of the stack trace; if loaded into memory, debugging is easier. But in these cases, this might not warrant the overhead.

conversion_service/converters/gis_converter/bootstrap/sql/filter/drop_and_recreate.sql
hixi: I think this was meant as a comment to what happens below, so I left it. But removing seems very sensible to me.

conversion_service/converters/gis_converter/extract/statistics/statistics.py
```diff
@@ -0,0 +1,261 @@
...
+                [
+                    element,
+                    '{}'.format(
+                        self._retrieve_statistic(table_name, where_clause_type=where_clause_type, **clause_kwargs)
+                    )
+                ]
+            )
+        query_results = sorted(query_results, key=lambda t: t[1], reverse=True)
+        if label:
+            [item.insert(0, label) for item in query_results]  # in place label addition
```
das-g: If you aren't going to use the resulting list, use a for-loop instead of a list comprehension.

conversion_service/converters/gis_converter/extract/statistics/osm_groups.py
hixi: YAML to the win - would be a good idea, but maybe rather at a later time (creating an issue for this?)

conversion_service/converters/gis_converter/extract/statistics/statistics.py
hixi: yes, this was the idea. Gathering can take very long, so if we can't write the file, it doesn't do unnecessary work.

conversion_service/converters/gis_converter/extract/statistics/osm_groups.py
das-g: > would be a good idea, but maybe rather at a later time

:+1:

:white_check_mark: issue filed: #75

conversion_service/converters/gis_converter/bootstrap/bootstrap.py
hixi: This isn't true when using `execute_psql` of course, so no reason to keep reading the files into memory at this point.

conversion_service/converters/gis_converter/helper/postgres_wrapper.py
```diff
@@ -0,0 +1,88 @@
...
+        return self._connection_parameters['dbname']
+
+    def get_user(self):
+        return self._connection_parameters['user']
+
+    def _get_connection(self, connection_parameters=None):
+        if connection_parameters is None:
+            connection_parameters = self._connection_parameters
+            psycopg2.connect(**connection_parameters)
+        return psycopg2.connect(**connection_parameters)
```
das-g: Why connect twice if no `connection_parameters` are passed?

conversion_service/converters/gis_converter/extract/statistics/statistics.py
```diff
@@ -0,0 +1,258 @@
...
+from converters.gis_converter.extract.statistics.osm_groups import *  # noqa: ignore * import
+from converters.gis_converter.helper.default_postgres import get_default_postgres_wrapper
+
+
+def gather_statistics(outfile):
+    with open(outfile, 'w') as out_file:
+        csv_writer = csv.writer(out_file, delimiter=';')
+        all_stats = Statistics().gather_all()
+        for stat in all_stats:
+            csv_writer.writerow(stat)
```
das-g: How about using [`csv_writer.writerows(all_stats)`](https://docs.python.org/3/library/csv.html#csv.csvwriter.writerows) and get rid of the explicit loop?

conversion_service/converters/gis_converter/extract/statistics/statistics.py
das-g: :white_check_mark: good idea, keep that order

conversion_service/converters/garmin_converter/to_garmin.py
das-g: :white_check_mark: `data_dir` is good. I see you already switched to that in e05c296.

conversion_service/converters/gis_converter/extract/statistics/statistics.py
das-g: :white_check_mark: done in 52830e8def85904aab308efcf662ad36dfd6197a

conversion_service/converters/gis_converter/bootstrap/bootstrap.py
das-g: :white_check_mark: I see you switched to my suggestion with da9eb4db9572c8a527b9e74a28022444ba08d6e2

conversion_service/converters/gis_converter/helper/postgres_wrapper.py
das-g: :white_check_mark: done with b90762f2325b5d624640906df02a60be00077bae, fixed a typo in 13074706db4af06bd6183465791fd3ea3c61102f

conversion_service/converters/gis_converter/helper/postgres_wrapper.py
hixi: oops...

das-g:
:+1: Good work!

And direction is good, too, as far as I can tell.

Feel free to merge once https://github.com/geometalab/osmaxx-conversion-service/pull/73#discussion_r48947248 and https://github.com/geometalab/osmaxx-conversion-service/pull/73#discussion_r48948490 are considered.

PR #92, opened 2016-01-22
Feature/split sql scripts

### LINKED ISSUE #91: Fix string to integer conversion

hixi:
Reviewed by:
- [x] @das-g 

Part of geometalab/osmaxx#334

(Accidentially) closes #91

Noteworthy: The new `cast_to_int_null_if_failed` function makes the sql-scripts more resilient against wrong type of values within the (text-only) osm-schema when converting to an integer.


conversion_service/converters/gis_converter/bootstrap/bootstrap.py
```diff
@@ -73,30 +73,34 @@ def _harmonize_database(self):
...
-        for filter_script in filter_sql_scripts_ordered:
-            filter_script_path = os.path.join(base_dir, filter_script)
+        for script_folder in filter_sql_script_folders:
+            script_folder_path = os.path.join(base_dir, script_folder)
             # FIXME: replace the drop/create commands so autocommit is not needed anymore!
-            self._postgres.execute_psycopg_file(filter_script_path, autocommit=True)
+            self._execute_sql_scripts_in_folder(script_folder_path, autocommit=True)
+
+    def _execute_sql_scripts_in_folder(self, folder_path, autocommit=False):
+        for script_path in sorted(glob.glob(folder_path + '/*.sql'), key=lambda folder: folder.split('/')[-1]):
```
das-g: I think [`os.path.basename`](https://docs.python.org/3/library/os.path.html#os.path.basename) already does what you want to achieve with your lambda function. Can we use `key=os.path.basename`?

das-g: Is there a reason not to use `os.path.join(folder_path, '*.sql')` for building the glob (the argument to `glob.glob(...)`)?

das-g: Pulling the first argument of `sorted(...)` out to a variable called `sql_scripts_in_folder` would make this easier to understand, I think.

hixi: I have rewritten this a bit, but I'm not sue, what you meant with the lambda; that is merely a sorting rule.

das-g: I meant
```python
    def _execute_sql_scripts_in_folder(self, folder_path, autocommit=False):
        sql_scripts_in_folder = glob.glob(os.path.join(folder_path, '*.sql'))
        for script_path in sorted(sql_scripts_in_folder, key=os.path.basename):
            # ...
```
(untested)

tests/converters/bootstrap_test.py
```diff
@@ -0,0 +1,79 @@
+from unittest import TestCase
+
+from unittest import mock
```
das-g: above two imports could be joined to
```python
from unittest import TestCase, mock
```

tests/converters/bootstrap_test.py
```diff
@@ -0,0 +1,79 @@
...
+                        ),
+                        autocommit=True
+                    ) for relative_script_path in expected_script_order
+                ]
+            )
+
+    def test_function_scripts_are_executed_in_correct_order(self, *args, **kwargs):
+        bootstrapper = bootstrap.BootStrapper(pbf_file_path=settings.OSMAXX_CONVERSION_SERVICE['PBF_PLANET_FILE_PATH'])
+        with mock.patch.object(bootstrapper, '_postgres') as postgres_mock:
+            assert bootstrapper._postgres is postgres_mock
```
das-g: Should this assertion be kept?

hixi: In this case it doesn't make sens anymore, you're right.

conversion_service/converters/gis_converter/bootstrap/bootstrap.py
das-g: to use my suggested variant, merge #93 (will be merged into `feature/split-sql-scripts`) before #92.

PR #95, opened 2016-01-25
Feature/#94 fix utf8 language issue

### LINKED ISSUE #94: Fix utf-8 issue in database

hixi:
Reviewed by:
- [x] @das-g 

Closes #94 
Closes geometalab/osmaxx-frontend#251


Dockerfile.worker
```diff
@@ -1,29 +1,63 @@
...
-MAINTAINER HSR Geometalab <geometalab@hsr.ch>
+# grab gosu for easy step-down from root
+RUN gpg --keyserver ha.pool.sks-keyservers.net --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates wget && rm -rf /var/lib/apt/lists/* \
+	&& wget -O /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/1.2/gosu-$(dpkg --print-architecture)" \
+	&& wget -O /usr/local/bin/gosu.asc "https://github.com/tianon/gosu/releases/download/1.2/gosu-$(dpkg --print-architecture).asc" \
+	&& gpg --verify /usr/local/bin/gosu.asc \
+	&& rm /usr/local/bin/gosu.asc \
+	&& chmod +x /usr/local/bin/gosu \
+	&& apt-get purge -y --auto-remove ca-certificates wget
```
das-g: I don't see `gosu` being used in this Dockerfile (or elsewere in this PR). Is it being used somewhere else?

Dockerfile.worker
```diff
@@ -1,29 +1,63 @@
...
-RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ trusty-pgdg main' $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
+RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ jessie-pgdg main' $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
+
+RUN apt-get update \
+	&& DEBIAN_FRONTEND=noninteractive apt-get install -y postgresql-common \
+	&& sed -ri 's/#(create_main_cluster) .*$/\1 = false/' /etc/postgresql-common/createcluster.conf \
+	&& DEBIAN_FRONTEND=noninteractive apt-get install -y\
+		postgresql-${PG_MAJOR} \
+		postgresql-contrib-${PG_MAJOR} \
+        postgresql-${PG_MAJOR}-postgis-${POSTGIS_MAJOR} \
```
das-g: Are the different indentation intentional?

Dockerfile.worker
```diff
@@ -1,29 +1,63 @@
...
 ENV PG_MAJOR 9.4
 ENV POSTGIS_MAJOR 2.1
 
-# postgres sources
-RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ trusty-pgdg main' $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
+RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ jessie-pgdg main' $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
+
+RUN apt-get update \
+	&& DEBIAN_FRONTEND=noninteractive apt-get install -y postgresql-common \
+	&& sed -ri 's/#(create_main_cluster) .*$/\1 = false/' /etc/postgresql-common/createcluster.conf \
```
das-g: :+1: elegant way to set that configuration

hixi: elegant because it isn't from me but from the postgres maintainers: https://github.com/docker-library/postgres/blob/762e07937086fb43082ab1523ac9d0191c2c347f/9.4/Dockerfile#L33

Dockerfile.worker
hixi: I should've used gosu, corrected now.

Dockerfile.worker
hixi: No, why should they be intentional? they stem from using different editors, should be remedied now.

Dockerfile.worker
das-g: Was a tabs/spaces issue. Converted everything to spaces with 7e4665aafee3c6617dc472c5aaad93c9a7477ded

Was visible to me, because I let GitHub display tabstops with the default width of 8.

Dockerfile.worker
das-g: Was a tabs/spaces issue. Converted everything to spaces with 7e4665aafee3c6617dc472c5aaad93c9a7477ded

Was visible to me, because I let GitHub display tabstops with the default width of 8.

Dockerfile.worker
das-g: :white_check_mark:

Dockerfile.worker
das-g: > Not sure if we should mention that as well in the Image...

Nah, that's trivial enough to be adopted without attribution.

Dockerfile.worker
```diff
@@ -53,50 +86,32 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
...
-  cd src &&\
-  git clone https://github.com/openstreetmap/osm2pgsql.git &&\
-  cd osm2pgsql &&\
-  mkdir build &&\
-  cd build &&\
-  cmake ..&&\
-  make &&\
-  make install
+    cd src &&\
+    GIT_SSL_NO_VERIFY=true git clone https://github.com/openstreetmap/osm2pgsql.git &&\
```
das-g: Is the certificate invalid or do we miss the CA in the docker image to verify it (or is there another issue that mandates `GIT_SSL_NO_VERIFY=true`)?

hixi: it just fails somehow, I'm guessing we're missing the CA, but the error did not say much, just ssl-verification failed...

Dockerfile.worker
```diff
@@ -116,12 +131,8 @@ RUN mkdir /etc/ssl/private-copy; mv /etc/ssl/private/* /etc/ssl/private-copy/; r
...
 COPY ./docker_entrypoint/worker $HOME/entrypoint
 
-USER postgres
-RUN service postgresql start && \
-    $HOME/entrypoint/00_make_db_utf8.sh && \
-    psql -f $HOME/entrypoint/01_activate_translit.sql
-
-USER root
+RUN gosu postgres service postgresql start && sleep 10 &&\
+    gosu postgres psql -f $HOME/entrypoint/01_activate_translit.sql
```
das-g: Consider putting the `sleep` invocation on a line of its own

Dockerfile.worker
```diff
@@ -1,28 +1,61 @@
...
 
 USER root
 
+# make the "en_US.UTF-8" locale so postgres will be utf-8 enabled by default
+RUN apt-get update && apt-get install -y apt-utils locales && rm -rf /var/lib/apt/lists/* \
+    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+
+ENV LANG en_US.utf8
+ENV LOCALE en_US.UTF-8
+ENV LC_ALL en_US.utf8
```
das-g: Are all these environment variables still required or is all that matters `--locale` option when creating the cluster?

hixi: Not sure, but I think none of them is really needed anymore. I will check and remove where possible.

Dockerfile.worker
hixi: Thanks for bringing this up, I almost forgot that we don't need it anymore ;-)

Long version: I did consider and let it there on purpose: it matches against the why this is needed in the first place. But it isn't needed anymore, because `service postgresql start` does return only after having the db up :-)

Issue #96, opened 2016-01-26
Refactor the service to just allow arbitrary polygons as area cut out


hixi:
part of / preparation for geometalab/osmaxx-frontend#329

- [x] add support for arbitrary multipolygons
- [x] deprecate old variants

hixi:
Closing, implemented in merged repos.

PR #106, opened 2016-01-29
Feature/#96 add arbitrary polygon support

hixi:
Part of #96 

Reviewed by:
- [x] @das-g


das-g:
I guess this should be reviewed (and if OK merged) without 79a3f94, i.e., only up to 7ca08cb?

hixi:
@das-g yes, the work-in-progress commit landed here in error, so removed.

conversion_service/converters/boundaries.py
```diff
@@ -59,7 +61,30 @@ def _get_cut_command(self, output_filename):
         ]
 
 
+class PolyfileCutter:
+    def __init__(self, poly_string):
+        utils.parse_poly_string(poly_string)
```
das-g: As far as I can see, [`parse_poly_string(...)`](https://github.com/geometalab/osmaxx-conversion-service/pull/106/files#diff-98706db1896bd60ddbe28477cf9a3a75R30) doesn't have any side effects (not even on the argument), so calling it without using the return value is moot.

conversion_service/converters/boundaries.py
```diff
@@ -59,7 +61,30 @@ def _get_cut_command(self, output_filename):
...
+        self.poly_string = poly_string
+
+    def cut_pbf(self, output_filename):
```
das-g: It's a bit strange that this takes the file name of the **out**put `*.pbf`, but not of the **in**put `*.pbf`. Adding that as an argument (preferably as the first argument after `self`) would make sense from a separation-of-concerns perspective, too: This class should not care whether the area(s) defined by the `*.poly` are cut from the planet or any other `*.pbf`, nor where those files can be found.

conversion_service/converters/boundaries.py
hixi: this might look like it, but it does some obscure magic: it checks, if the polyfile-line-string-or-whatever-we-want-to-call-it is a valid polygon (this raises an error, as in the [test here](https://github.com/geometalab/osmaxx-conversion-service/pull/106/files#diff-343e4ff9826a1d1b5901def0284ad656R81)). But this is a bit of a bad implementation for such a check, do you have a suggestion for a short, comprehensive, understandable replacement? ;-)

tests/clipping_polygon/to_polyfile_test.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class CreatePolyfileStringTest(TestCase):
+    def setUp(self):
+        self.poly_1 = Polygon(((0, 0), (0, 1), (1, 1), (0, 0)))
+        self.poly_2 = Polygon(((1, 1), (1, 2), (2, 2), (1, 1)))
+        self.multi_polygon_1 = MultiPolygon(self.poly_1, self.poly_2)
+        self.multi_polygon_2 = MultiPolygon([self.poly_1, self.poly_2])
+
+        self.holy_poly_3 = Polygon(((0, 0), (0, 1), (1, 1), (0, 0)), ((1, 1), (1, 2), (2, 2), (1, 1)))
```
das-g: Sorry to ruin [the joke](http://tvtropes.org/pmwiki/pmwiki.php/Main/HolyHandGrenade), but I'm confident you mean "holey", not "holy". :stuck_out_tongue_winking_eye:

tests/clipping_polygon/to_polyfile_test.py
```diff
@@ -0,0 +1,99 @@
...
+        self.multi_polygon_3 = MultiPolygon(self.holy_poly_3)
+
+    def test_create_poly_file_string_when_valid_multipolygon_returns_correct_string(self):
```
das-g: To avoid a similar confusion we had with the following test's name, consider calling this one here `test_create_poly_file_string_when_multipolygon_is_valid_returns_correct_string`.

tests/clipping_polygon/to_polyfile_test.py
hixi: No, I am absolutely aware of the correct spelling, that is supposed to be part of that joke...

tests/clipping_polygon/to_polyfile_test.py
```diff
@@ -0,0 +1,99 @@
...
+            '  2.000000E+00 2.000000E+00',
+            '  1.000000E+00 1.000000E+00',
+            'END',
+            'END',
+            '',
+        ])
+        result = create_poly_file_string(self.multi_polygon_3)
+        self.assertMultiLineEqual(expected, result)
+
+    def test_create_poly_file_string_when_invalid_input_raises_type_error(self):
```
das-g: `..._when_input_is_invalid_raise_...`

tests/clipping_polygon/to_polyfile_test.py
```diff
@@ -0,0 +1,99 @@
...
+        no_multi_polygon_at_all = 'asda'
+        self.assertRaises(TypeError, create_poly_file_string, no_multi_polygon_at_all)
+
+    def test_create_poly_file_string_when_invalid_geometry_raises_type_error(self):
```
das-g: `..._when_geometry_is_invalid_raises_...`

conversion_service/converters/boundaries.py
hixi: `Information Expert`: The only thing that needs to be controlled from the outside is the output pbf filename, all other details are internals.
I can be convinced that the _name_ of this class/function is chosen badly, but nobody else knows and should need to know, where the input pbf file lies, we're only interested in the output.

tests/clipping_polygon/to_polyfile_test.py
hixi: :heavy_check_mark:

conversion_service/converters/boundaries.py
das-g: I guess you mean [`TestPolyfileCutter.test_init_when_invalid_poly_raises_error`](https://github.com/geometalab/osmaxx-conversion-service/pull/106/files#diff-5371d8df004b5ad831d8d63540720d20R120), not `CreatePolyfileStringTest.test_create_poly_file_string_when_invalid_geometry_raises_type_error`. Yeah, I overlooked that. Maybe this warrants a comment in the code?
```python
        utils.parse_poly_string(poly_string)  # Assume poly_string is valid if parsing doesn't cause an Exception.
```
or
```python
        # Parse the poly_string to cause an Exception if it is invalid.
        utils.parse_poly_string(poly_string)  # We discard the return value, as we're only interested in the side effect.
```

conversion_service/converters/boundaries.py
das-g: Would it be over-engineering to introduce a `PolyfilePlanetCutter` class that wraps a generic `PolyfilePbfCutter`?

conversion_service/converters/boundaries.py
hixi: Adding a comment is good, even though I'd like self documenting code better... but comment for now works :-)

conversion_service/converters/boundaries.py
hixi: Yes, I believe that it would be over-engineering, especially since we don't need a generalized class - if we do, that can still be achieved quickly and easily (we're not blocking future development).

conversion_service/converters/boundaries.py
```diff
@@ -59,7 +62,57 @@ def _get_cut_command(self, output_filename):
...
+    def __init__(self, osmosis_polygon_file_content):
+        """
+        Context Manager that returns an pbf file path cutted along the string representation of a osmosis polygon file.
+
+        Args:
+            osmosis_polygon_file_content: string representation of the polygon file.
+
+        Returns:
+            the path to the pbf file usable during context
+        """
```
das-g: I think this should be the class's docstring, not `__init__`'s. (`__init__` returns None, not a path.)

PR #110, opened 2016-02-02
Feature/#96 add arbitrary polygon support

hixi:
Part of #96 

Adds support for an clipping area.

Reviewed by
- [x] @das-g

### DIFF
--- a/Dockerfile.rest_api
+++ b/Dockerfile.rest_api
@@ -1,4 +1,6 @@
-FROM python:3.4
+FROM geometalab/gdal-debian
+
+USER root
 
 MAINTAINER HSR Geometalab <geometalab@hsr.ch>
 
@@ -6,7 +8,8 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y\
 \
     libgeos-dev \
-    libgeos++-dev
+    libgeos++-dev \
+    python3-pip
 
 # Install required Python packages:
 ENV HOME /home/py
@@ -15,7 +18,7 @@ WORKDIR $HOME
 
 ADD requirements.txt $HOME/
 
-RUN pip install honcho
+RUN pip3 install honcho
 
 RUN pip3 install -r requirements.txt
 
@@ -29,4 +32,3 @@ RUN mkdir -p $HOME/docker_entrypoint/conversion_service $HOME/entrypoint
 COPY ./docker_entrypoint/conversion_service $HOME/entrypoint
 
 ENTRYPOINT ["/home/py/entrypoint/entrypoint.sh"]
-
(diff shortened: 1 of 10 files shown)


Dockerfile.rest_api
```diff
@@ -15,7 +18,7 @@ WORKDIR $HOME
 
 ADD requirements.txt $HOME/
 
-RUN pip install honcho
+RUN pip3 install honcho
```
hixi: switched to python3 so we don't need to install another python version (besides the standard python3).

Dockerfile.rest_api
```diff
@@ -6,7 +8,8 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y\
 \
     libgeos-dev \
-    libgeos++-dev
+    libgeos++-dev \
```
hixi: seems top be a dependency for the newer version of `djangorestframework-gis`, therefore included.

conversion_service/clipping_geometry/serializers.py
```diff
@@ -0,0 +1,27 @@
...
+    def validate_clipping_multi_polygon(self, clipping_multi_polygon):
+        if type(clipping_multi_polygon) != geos.MultiPolygon:
+            raise serializers.ValidationError(
+                _(
+                    "The received geometry is not a MultiPolygon. Received type of {other_type}.".format(
+                        other_type=clipping_multi_polygon.__class__.__name__
+                    )
+                )
+            )
+        if clipping_multi_polygon.num_coords <= 0:
```
hixi: if it has not more than one coordinate, it's not valid, but the geoJSON to python conversion doesn't catch this and the error is raised on database level (integrity error), which doesn't leave much information, why it happened.

Dockerfile.rest_api
```diff
@@ -6,7 +8,8 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
...
+    python3-pip
```
hixi: other image, apparently doesn't include python3-pip

conversion_service/clipping_geometry/serializers.py
```diff
@@ -0,0 +1,29 @@
...
+                    "The received geometry is not a MultiPolygon. Received type of {other_type}.".format(
+                        other_type=clipping_multi_polygon.__class__.__name__
+                    )
+                )
+            )
+        # prevent obscure IntegrityError in the database:
+        # occurs for example when the geoJSON seems valid, but contains non-coordinate string(s).
+        if clipping_multi_polygon.num_coords <= 0:
+            raise serializers.ValidationError(
+                _("Invalid coordinates: Expecting at least one coordinate set, received none.")
```
das-g: 'set' might induce the wrong association with the concept of a mathematical set (an unordered set). How about 'coordinate tuple', or &ndash; as we're only dealing with 2D &ndash; 'coordinate pair'?

hixi: good idea :-)

PR #111, opened 2016-02-02
Feature/#108 make version app work again

hixi:
Reviewed by:

- [x] @das-g 

No test written for this small feature, tested manually. Reasoning: This app is very unlikely to change in the future and completely orthogonal to the rest of the project (no inter-dependencies).

### DIFF
--- a/conversion_service/version/views.py
+++ b/conversion_service/version/views.py
@@ -1,0 +1,12 @@
+import importlib
+
+from django.conf import settings
+from django.shortcuts import render_to_response
+
+
+def show_version_number(request):
+    version = importlib.import_module('__init__').__version__
+    return render_to_response(template_name='version/show_version.html', context={
+        'version_number': version,
+        'DEBUG': settings.DEBUG,
+    })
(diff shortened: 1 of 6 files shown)


conversion_service/version/views.py
```diff
@@ -0,0 +1,12 @@
...
+
+from django.conf import settings
+from django.shortcuts import render_to_response
+
+
+def show_version_number(request):
+    version = importlib.import_module('__init__').__version__
+    return render_to_response(template_name='version/show_version.html', context={
+        'version_number': version,
+        'DEBUG': settings.DEBUG,
```
das-g: Rather than displaying the state of the `DEBUG` setting, we should probably check whether we _actually_ are in DEBUG mode (which also depends on whether the request’s IP address (`request.META['REMOTE_ADDR']`) is in the `INTERNAL_IPS` setting), simply by using the template variable `debug` [that Django already sets for us](https://docs.djangoproject.com/en/1.8/ref/templates/api/#django-template-context-processors-debug) (so we could just use it in the template without passing it here).

das-g: :white_check_mark: leave as is

@hixi convinced me that the DEBUG _setting_ **is** what we want to display here.

PR #119, opened 2016-02-18
Feature/#96 arbitrary polygon support

hixi:
**Do not MERGE** just yet.

- [x] remove  dc0a3ec again

Implements the conversion-service side of this feature. Also features a partial rewrite, maybe use the test commits to review especially commit 472c017.

Reviewed by:
- [x] @das-g


osmaxx_conversion_service/requirements.txt
```diff
@@ -13,6 +13,7 @@ django-extensions==1.5.9
 Werkzeug==0.10.4
```
das-g: the `pip` command at https://github.com/geometalab/osmaxx-conversion-service/blob/feature/%2396_arbitrary_polygon_support/docs/index.md#testing is out-of-date: Should probably be
```bash
pip install -r osmaxx_conversion_service/requirements.txt
```
instead of
```bash
pip install -r requirements.txt
```

das-g:
Unrelated to this change, but `setup.py` has a bogus email address. Should probably use the `geometalab` one instead?

osmaxx/conversion/_settings.py
```diff
@@ -1,16 +1,17 @@
+from datetime import timedelta
 from django.conf import settings
 
-OSMAXX_CONVERSION_SERVICE = {
+CONVERSION_SETTINGS = {
+    'result_harvest_interval': timedelta(minutes=1).total_seconds(),
```
das-g: Either include 'seconds' in this setting's name, or postpone the conversion to seconds to the setting user `result_harvester.Command.handle(...)`.

osmaxx/conversion/_settings.py
```diff
@@ -1,16 +1,17 @@
...
     'PBF_PLANET_FILE_PATH': '/var/data/osm-planet/planet-latest.osm.pbf',
-    'RESULT_DIR': '/tmp/osmaxx-conversion-service/results',
     'RESULT_TTL': -1,  # never expire!
 }
 
 if hasattr(settings, 'OSMAXX_CONVERSION_SERVICE'):
-    OSMAXX_CONVERSION_SERVICE.update(settings.OSMAXX_CONVERSION_SERVICE)
+    CONVERSION_SETTINGS.update(settings.OSMAXX_CONVERSION_SERVICE)
```
das-g: Is the name `OSMAXX_CONVERSION_SERVICE` still correct here?

osmaxx_conversion_service/requirements.txt
hixi: yes, but.... thinking a bit about it, shouldn't the requirements.txt be inside `<repo_root>/` as before, because setup.py is there as well? Those aren't the requirements for the omaxx_conversion_service (ie django project), but for _all_ osmaxx.

osmaxx/conversion/converters/converter.py
```diff
@@ -0,0 +1,77 @@
...
+            out_srs=None
+    ):
+        self._conversion_format = conversion_format
+        self._output_zip_file_path = output_zip_file_path
+        self._area_name = area_name
+        self._polyfile_string = osmosis_polygon_file_string
+        self._name_prefix = filename_prefix
+        self._out_srs = out_srs
+
+        _format_process = {
```
das-g: This variable could live on the class instead of the instance.

hixi:
good catch, will do a PR.

osmaxx/conversion/converters/converter.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(
+            self,
+            *,
+            conversion_format,
+            area_name,
+            osmosis_polygon_file_string,
+            output_zip_file_path,
+            filename_prefix,
+            out_srs=None
+    ):
```
das-g: Those are a lot of arguments for a single class's initializer. Should these be split up, so the signature becomes

```python
area = Area(
    name='some name',
    osmosis_polygon_file_string="C:\My Files\my.poly",
)
conversion_output_options = ConversionOutputOptions(
    format=FGDB,
    output_zip_file_path="D:\output.zip",
    filename_prefix='foobar',
    srs=EPSG_5000,
)
conversion = Conversion(area=area, output_options=conversion_output_options)
```
?

osmaxx_conversion_service/requirements.txt
das-g: Fine with me.

osmaxx/conversion/converters/converter_gis/gis.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class GISConverter:
+    def __init__(self, *, conversion_format, out_zip_file_path, base_file_name, out_srs, polyfile_string):
+        """
+        Converts a specified pbf into the specified format.
+
+        Args:
+            out_zip_file_path: path to where the zipped result should be stored, directory must already exist
+            conversion_format: One of ['fgdb', 'shapefile', 'gpkg', 'spatialite']
```
das-g: I don't think human readers need square brackets

osmaxx/conversion/converters/converter_gis/gis.py
```diff
@@ -0,0 +1,53 @@
...
+
+from osmaxx.conversion.converters.boundaries import pbf_area
+from osmaxx.conversion.converters.converter_gis.bootstrap import bootstrap
+from osmaxx.conversion.converters.converter_gis.extract.db_to_format.extract import extract_to
+from osmaxx.conversion.converters.converter_gis.extract.statistics.statistics import gather_statistics
+from osmaxx.conversion.converters.utils import zip_folders_relative
+
+
+class GISConverter:
+    def __init__(self, *, conversion_format, out_zip_file_path, base_file_name, out_srs, polyfile_string):
```
das-g: might also be a candidate for packing arguments into objects

osmaxx/conversion/converters/converter_gis/extract/db_to_format/extract.py
```diff
@@ -2,25 +2,26 @@
...
-from osmaxx.converters.converter_settings import OSMAXX_CONVERSION_SERVICE
+from osmaxx.conversion._settings import CONVERSION_SETTINGS
+from osmaxx.conversion.formats import FGDB, SHAPEFILE, GPKG, SPATIALITE
 
 FORMATS = {
-    'fgdb': {
+    FGDB: {
         'ogr_name': 'FileGDB',
         'extension': '.gdb',
         'extraction_options': [],
```
das-g: Could we include this information in `osmaxx.conversion.formats.FORMAT_DEFINITIONS` or `osmaxx.conversion.formats.FGDB` ... ?

osmaxx/conversion/formats.py
```diff
@@ -0,0 +1,80 @@
+import uuid
+
+from django.utils.translation import gettext as _
+
+FGDB, SHAPEFILE, GPKG, SPATIALITE, GARMIN = 'fgdb', 'shapefile', 'gpkg', 'spatialite', 'garmin'
```
das-g: Could we store the actual OutputFormat objects in these constants instead of strings?

osmaxx/conversion/converters/converter.py
```diff
@@ -0,0 +1,77 @@
...
+        self._name_prefix = filename_prefix
+        self._out_srs = out_srs
+
+        _format_process = {
+            GARMIN: self._create_garmin_export,
+            FGDB: self._extract_postgis_format,
+            SHAPEFILE: self._extract_postgis_format,
+            GPKG: self._extract_postgis_format,
+            SPATIALITE: self._extract_postgis_format,
+        }
```
das-g: Would it make sense to include this information in `osmaxx.conversion.formats.FORMAT_DEFINITIONS`?

osmaxx/conversion/converters/converter.py
das-g: Oops, no, it can't (so easily), due to dependency on `self`.

osmaxx/conversion/management/commands/result_harvester.py
```diff
@@ -0,0 +1,76 @@
...
+        queue = django_rq.get_queue()
+        active_jobs = conversion_models.Job.objects.exclude(status__in=conversion_models.Job.STATUSES_FINAL)\
+            .values_list('rq_job_id', flat=True)
+        for job_id in active_jobs:
+            self._update_job(job_id=job_id, queue=queue)
+
+    def _update_job(self, job_id, queue):
+        job = queue.fetch_job(job_id)
+        if job is None:  # already processed by someone else
+            return
```
das-g: Shouldn't a final status be assured in that case?

osmaxx/conversion/management/commands/result_harvester.py
```diff
@@ -0,0 +1,76 @@
...
+        except:
+            logger.error('failed to send notification for job {}'.format(conversion_job.id))
+            pass
+
+
+def add_file_to_job(*, conversion_job, result_zip_file):
+    conversion_job.resulting_file.name = conversion_job.zip_file_relative_path()
+    new_path = os.path.join(settings.MEDIA_ROOT, conversion_job.resulting_file.name)
+    if not os.path.exists(os.path.dirname(new_path)):
+        os.makedirs(os.path.dirname(new_path), exist_ok=True)
```
das-g: consider extracting `os.path.dirname(new_path)` (suggested name: `parent_dir` or `target_dir`)

osmaxx/conversion/viewsets.py
```diff
@@ -0,0 +1,38 @@
...
+        super().perform_create(serializer=serializer)
+        serializer.instance.start_conversion()
+
+    @detail_route(methods=['get'], permission_classes=[permissions.IsAuthenticated], url_path='download-zip')
+    def download_zip(self, request, pk):
+        try:
+            job = self.queryset.get(pk=pk)
+        except ObjectDoesNotExist:
+            return Response(status=404)
+        if not job.resulting_file.name:
```
das-g: file present, but has no name ... or what?

osmaxx_conversion_service/Procfile.mediator.dev
```diff
@@ -1 +1,2 @@
 mediator: python3 ./osmaxx_conversion_service/manage.py runserver_plus ${APP_HOST}:${APP_PORT}
+the_collector: python3 ./osmaxx_conversion_service/manage.py result_harvester
```
das-g: can we drop the `the_`?

osmaxx_conversion_service/Procfile.mediator.prod
```diff
@@ -1 +1,2 @@
-api: gunicorn --workers ${NUM_WORKERS} --chdir=osmaxx_conversion_service config.wsgi --bind ${APP_HOST}:${APP_PORT}
+mediator: gunicorn --workers ${NUM_WORKERS} --chdir=osmaxx_conversion_service config.wsgi --bind ${APP_HOST}:${APP_PORT}
+the_collector: python3 ./osmaxx_conversion_service/manage.py result_harvester
```
das-g: can we drop the `the_`?

tests/clipping_area/model_test.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+
+
+@pytest.mark.django_db()
+def test_osmosis_polygon_file_string_property_returns_osmosis_polygon_file(valid_clipping_area):
+    assert valid_clipping_area.osmosis_polygon_file_string is not None
+    assert valid_clipping_area.osmosis_polygon_file_string != ''
```
das-g: Are you just testing the fixture here? If so, put the assertions into the fixture code instead of into a test.

tests/clipping_area/model_test.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+@pytest.mark.django_db()
+def test_representation(valid_clipping_area):
+    assert str(valid_clipping_area) is not None
+    assert str(valid_clipping_area) == "test (1)"
```
das-g: The second assertion implies the first. Is the first one still useful?

tests/clipping_area/viewset_test.py
```diff
@@ -164,3 +167,30 @@ def test_detail_view_without_login_fails(self):
...
+def test_clipping_area_creation_succeeds(api_client_authenticated, clipping_area_hsr_data):
+    response = api_client_authenticated.post(reverse('clipping_area-list'), clipping_area_hsr_data, format='json')
+    assert response.status_code == 201
+    expected_data = clipping_area_hsr_data.copy()
+    expected_data['id'] = 1
+    assert response.json() == expected_data
+
+
+@pytest.mark.django_db()
+def test_clipping_area_creation_fails(api_client, clipping_area_hsr_data):
```
das-g: Test name should probably hint at the fact that this is without authentication?

tests/clipping_area/viewset_test.py
```diff
@@ -164,3 +167,30 @@ def test_detail_view_without_login_fails(self):
...
+    response = api_client.post(reverse('clipping_area-list'), clipping_area_hsr_data, format='json')
+    assert response.status_code == 403
+
+
+@pytest.mark.django_db()
+def test_clipping_area_detail_access_succeeds(api_client_authenticated, valid_clipping_area):
```
das-g: If `valid_clipping_area` is an already persisted one, the fixture name should probably indicate that.

tests/clipping_area/viewset_test.py
```diff
@@ -164,3 +167,30 @@ def test_detail_view_without_login_fails(self):
...
+    response = api_client_authenticated.get(reverse('clipping_area-detail', args=[valid_clipping_area.id]))
+    assert response.status_code == 200
+
+
+@pytest.mark.django_db()
+def test_clipping_area_detail_access_fails(api_client, valid_clipping_area):
```
das-g: Test name should probably hint at the fact that this is without authentication?

tests/conftest.py
```diff
@@ -89,3 +90,60 @@ def pytest_configure():
...
+
+
+@pytest.fixture
+def api_client():
+    from rest_framework.test import APIClient
+    return APIClient()
+
+
+@pytest.fixture
+def api_client_authenticated(api_client):
```
das-g: to follow the pattern of the above fixtures, I'd name this one `authenticated_api_client`.

tests/conftest.py
```diff
@@ -89,3 +90,60 @@ def pytest_configure():
...
+    mark your test with @pytest.mark.django_db()
+
+
+    Args:
+        api_client: api-client fixture
+
+    Returns:
+        Authenticated Client
+    """
+    return authenticated_client(api_client)
```
das-g: heh, clever

tests/conversion/conftest.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+@pytest.fixture(params=format_list)
+def conversion_format(request):
+    return request.param
+
+
+@pytest.fixture
+def area_name():
+    return 'area_name'
```
das-g: How about 'Neverland'?

tests/conversion/conftest.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+@pytest.fixture
+def conversion_job(conversion_parametrization, server_url):
+    from osmaxx.conversion.models import Job
+    return Job.objects.create(own_base_url=server_url, parametrization=conversion_parametrization)
+
+
+@pytest.fixture
+def conversion_job_started(conversion_parametrization, server_url, fake_rq_id):
```
das-g: how about naming this `started_conversion_job`?

tests/conversion/conftest.py
```diff
@@ -0,0 +1,144 @@
...
+    from osmaxx.conversion.models import Job
+    return Job.objects.create(own_base_url=server_url, parametrization=conversion_parametrization, rq_job_id=fake_rq_id, status=Job.STARTED)
+
+
+@pytest.fixture
+def conversion_job_failed(conversion_parametrization, server_url, fake_rq_id):
```
das-g: how about naming this `failed_conversion_job`?

tests/conversion/conftest.py
```diff
@@ -0,0 +1,144 @@
...
+    from osmaxx.conversion.models import Job
+    return Job.objects.create(own_base_url=server_url, parametrization=conversion_parametrization, rq_job_id=fake_rq_id, status=Job.FAILED)
+
+
+@pytest.fixture
+def conversion_job_finished(request, conversion_parametrization, server_url, fake_rq_id, empty_zip):
```
das-g: how about naming this `finished_conversion_job`?

tests/conversion/view_test.py
```diff
@@ -0,0 +1,110 @@
...
+def test_conversion_job_absolute_url_resolves_correct(conversion_job, server_url):
+    url = server_url + reverse('conversion_job-detail', kwargs={'pk': conversion_job.id})
+    assert conversion_job.get_absolute_url() == url
+
+
+@pytest.mark.django_db()
+def test_conversion_job_creation_enqueues(api_client_authenticated, conversion_job_data, mocker):
+    def rq_enqueue_with_settings_mock_return():
+        FakeRQ = namedtuple('FakeRQ', 'id')
+        return FakeRQ(id=42)
```
das-g: Used twice. Candidate for another fixture?

osmaxx/conversion/_settings.py
hixi: I believe so, because from a user perspective (aka "django configuration") this concerns the entire (blackbox) of the conversion service.

osmaxx/conversion/converters/converter.py
hixi: I agree it is a bit obscure, but I think this is the best solution non-the-less...

osmaxx/conversion/converters/converter.py
hixi: Two things that I see that speak against using data-classes (which I think are an anti-pattern per se)

Serialization: we obscure the process of serialization inside the redis-queue (they stay strings and can be serialized without issues but it's is using pickle/unpickle for that, so you would have a pickled object inside another - not so nice)

Indirection: even though there are a lot of parameters, they are explicit and hiding them doesn't sound like a good idea to me. (Also: we have to read more code to get the direct information, eg. of what does the area consist?)

As long as the passed objects don't convey more than just encapsulated data I'm against doing this. It only enhances readability in the first instance, when you then have to look inside the data-objects, IMHO it gets more convoluted to understand what is conveyed and therefore used.

osmaxx/conversion/converters/converter_gis/gis.py
hixi: See comment [above](https://github.com/geometalab/osmaxx-conversion-service/pull/119#discussion_r53461258)

osmaxx/conversion/converters/converter_gis/extract/db_to_format/extract.py
hixi: Why? This (ogr/file/disk related) information is not used anywhere else than in this module, so I think this is the right place. IMHO this is keeping the information details nicely separated.

osmaxx/conversion/formats.py
hixi: These is the common ground for all modules, the technical identifiers. All other information is specific to their respective modules. The biggest benefit I see is that using this is very straight-forward (no dictionary keys that have to be know, just plain and simple constants).

Also see comment [above](https://github.com/geometalab/osmaxx-conversion-service/pull/119#discussion_r53461753).

osmaxx/conversion/converters/converter.py
hixi: I don't believe so, no. The conversion should be the one who knows what to do with it's received information, and we shouldn't pass functions[1] for serialization reasons.

[1]: if pass function names, we're again in the same position as now.

osmaxx/conversion/management/commands/result_harvester.py
hixi: And what would you do in that case? Through an error? Whatever is gone is gone at this point. If we wanted, we could _add_ a final state if no such is set, but anything else is just guess work, since we cannot extract any information at this point from the worker/queue, since it is, well, gone ;-)

osmaxx/conversion/viewsets.py
hixi: resulting_file is always not None, but "name" (actually the 'path') is None when no file is appended to the file-field.

-> worth a comment here.

osmaxx_conversion_service/Procfile.mediator.prod
hixi: should be `harvester` anyway.

osmaxx_conversion_service/Procfile.mediator.dev
hixi: should be `harvester` anyway.

tests/clipping_area/model_test.py
hixi: Yes, indeed. I need to keep a dummy test to assure the fixture is being used though.

tests/clipping_area/model_test.py
hixi: no, the same is true for the assert above.

tests/clipping_area/viewset_test.py
hixi: good catch, thanks.

tests/clipping_area/viewset_test.py
hixi: good catch, didn't think of that, thanks.

tests/clipping_area/viewset_test.py
hixi: :+1:

tests/conftest.py
hixi: :+1:

tests/conversion/conftest.py
hixi: Middle Earth has a space - very good Idea :-)

tests/conversion/conftest.py
hixi: looking at the examples above, I think that makes a lot of sense :-)

tests/conversion/conftest.py
hixi: :+1:

tests/conversion/view_test.py
hixi: good catch, didn't notice that one before!

osmaxx/conversion/converters/converter_gis/extract/db_to_format/extract.py
das-g: > Why?

Because something here screams "S-Pattern" at me. (**s**trategy pattern or **s**tate pattern. More the former here.)

But maybe my feeling is misguided in that regard.

osmaxx/conversion/management/commands/result_harvester.py
das-g: If it's gone from the queue but not in a final state on the DB, that'd at least warrant a warning in the log, wouldn't it?

tests/clipping_area/model_test.py
das-g: If it ever becomes otherwise unused, shouldn't it be removed at that point? (Of course, if we don't observe coverage closely, we might not notice.)

tests/clipping_area/model_test.py
das-g: > the same is true for the assert above

Nope, see comment https://github.com/geometalab/osmaxx-conversion-service/commit/0bffa65ecf2cd61234f442aa376c0211615d602a#commitcomment-16363139

das-g:
Oh no,  dc0a3ec has been re-introduced by 6c82c4f1d225a06c874dc5c73038d5ffb4e5977f

das-g:
Suggestions:
### With history rewriting
I have made a merge-preserving rebase removing  dc0a3ec and 6c82c4f and pushed that to [my fork](https://github.com/das-g/osmaxx-conversion-service/tree/feature/%2396_arbitrary_polygon_support).

So we could `git reset` `feature/#96_arbitrary_polygon_support` (current tip 2bcb163) to [`das-g/osmaxx-conversion-service/feature/#96_arbitrary_polygon_support`](https://github.com/das-g/osmaxx-conversion-service/commits/feature/%2396_arbitrary_polygon_support) (current tip adb155b170470217773aae1a15bf85fcf8efe3ff). Any local changes that are descendants of dc0a3ec would have to be rebased.

### Without history rewriting
Make a new commit that reverts dc0a3ec.

das-g:
Talked to @hixi: We'll rectify this without history rewriting.

hixi:
**Do not MERGE** just yet.

hixi:
This has since been integrated in the other repository, geometalab/osmaxx-frontend#475 is the corresponding pull request. Closing.

PR #128, opened 2016-03-03
Feature/enhance tests

hixi:
Reviewed by:

- [x] @das-g
- [x] @hixi

Adds the very slow sql tests also to travis and allows to run tests more easily using the Makefile.


docs/index.md
```diff
@@ -126,30 +126,40 @@ Using the instruction below for testing, the flake8 tool is also run.
...
-The most simple way is to run an instance using docker (in a separate bash terminal):
+Install testing requirements.
 
 ```bash
-docker run -p 6379:6379 --rm --name redis-local redis
+$ pip install -r requirements.txt
 ```
 
-Install testing requirements.
+Run with runtests, excluding the very slow (>30 minutes) tests:
```
das-g: What does "[r]un with runtests [...]" mean here?
- "run with `runtests.py`
- "run with the `runtests-*` `make` target"
- or just "run the tests"
- ... something else

docs/index.md
```diff
@@ -126,30 +126,40 @@ Using the instruction below for testing, the flake8 tool is also run.
...
-$ tox
+$ make tests
+# or alternatively
+$ make runtests-slow
 ```
 
+To clean up all after the tests, you can use
+
+```bash
+$ make clean
```
das-g: Would it be OK to get rid of the `$` prompt indicator in the documentation? It makes copy-pasting harder.

(According to `git grep --extended-regexp '(^|\s)\$\s'`, it only seems to exist in this file here, `docs/index.md`)

docs/index.md
```diff
@@ -126,30 +126,40 @@ Using the instruction below for testing, the flake8 tool is also run.
...
-The most simple way is to run an instance using docker (in a separate bash terminal):
+Install testing requirements.
 
 ```bash
-docker run -p 6379:6379 --rm --name redis-local redis
+$ pip install -r requirements.txt
 ```
 
-Install testing requirements.
+Run with runtests, excluding the slow (&sum; > 1 minute) tests:
```
das-g: What does "[r]un with runtests [...]" mean here?
- "run with `./runtests.py`
- "run with the `runtests-*` `make` target"
- or just "run the tests"
- ... something else

docs/index.md
hixi: sure!

docs/index.md
hixi: run the tests with make, which uses the runtests.py. Clearer now?

